PR #11, opened 2020-09-29
speed benchmark

svlandeg:
Initial draft for speed benchmark, using spaCy, stanza, flairNLP, HF transformers, etc.

- TODO: define default corpus? Currently only works after dropping in .txt files in `texts`.
- TODO: double-check whether some of the other libraries have better/faster capabilities for multi-doc processing

Also updated the readme and config files from other projects, according to the latest changes.


svlandeg:
Ah, the CI fails because the updates to the configs etc are not yet reflected in a new `spacy-nightly` release. I think we can probably merge this anyway, and update pins & releases etc later.

PR #18, opened 2020-10-09
WIP update projects

svlandeg:
Updates to projects after recent `spacy-nightly` refactors.
Trying to get the CI go back to green.


ines:
I think we shouldn't really have `spacy-nightly` itself as a requirement anywhere – that's likely just going to cause us problems down the line because the projects themselves are executed via spaCy already (and we now have a mechanism for specifying version dependencies if we need to: by setting `spacy_version` to a version range in the `project.yml`).

In the FastAPI project, the requirement was needed for the Docker setup, I think. We should probably add a note about that in the `requirements.txt`.

svlandeg:
Quick status update:
- I think the failure for the "speed" project on Windows should be fixed with the encoding fix in https://github.com/explosion/spaCy/pull/6234
- The FastAPI & Emerson projects will fail until we've released new spacy-nightly models with the "strings.json" serialized

ines:
Update: `test_textcat_goemotions_project` should pass with the latest small fix I made to spaCy.

svlandeg:
Yay, green! 
I reviewed once more, and looks good to me.

PR #20, opened 2020-10-16
Fix train gpu_id in textcat_goemotions project

adrianeboyd:


svlandeg (commented):
We should probably also remove `gpu = -1` from the `system` block in `cnn.cfg`, as this is not used and can be confusing...

PR #25, opened 2021-01-14
Add corpus path as DocBin() will fail if doesnt exist

gonzobrandon:
clean install Linux and Mac error thrown as corpus path not in repo.

### DIFF
--- a/tutorials/nel_emerson/project.yml
+++ b/tutorials/nel_emerson/project.yml
@@ -71,7 +71,7 @@ commands:
   - name: corpus
     help: "Create a training and dev set from the manually annotated data"
     script:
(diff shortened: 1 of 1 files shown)


svlandeg:
Hm, I see your point, but I thought that would have been taken care off by putting `corpus` in the list of directories the project will automatically create?
https://github.com/explosion/projects/blob/778eb2dabce5213b639f6b8d6ad52ae31a87899c/tutorials/nel_emerson/project.yml#L18

Is that not working correctly?

gonzobrandon:
You are right. I made the mistake of `cd` to the repo. directory. All is working with the clone

PR #26, opened 2021-01-18
Example SpanCategorizer project

honnibal:
The project I've been using to develop the spancat. Needs Thinc 456 and spaCy 6747.


svlandeg:
Ah I see the default config was originally overridden with custom values. I just ran `init-config` again and committed the changes. I'll have a look tomorrow and benchmark with the original configs - perhaps we should keep those and remove the `init-config` command from the `project.yml`.

svlandeg:
While this can certainly be improved further, the current version works with the current spaCy implementation, and could be useful to demonstrate this new experimental feature :-)

adrianeboyd:
I tried it out and it ran fine, only the evaluate step doesn't show anything currently.

I think it would be a good idea to add `spacy_version` to `project.yml` so users with v3.0 see a warning.

Do you think there should be a test for this project, or maybe experimental projects should usually be skipped?

svlandeg:
> 
> I tried it out and it ran fine, only the evaluate step doesn't show anything currently.

Yep, I know, there's a TODO note in the .yml and the corresponding Ora ticket.

> 
> I think it would be a good idea to add `spacy_version` to `project.yml` so users with v3.0 see a warning.
> 
Ah, yes, we should do that.

> Do you think there should be a test for this project, or maybe experimental projects should usually be skipped?
I'll add one, it should still run.

svlandeg:
(the test will obviously fail as long as spaCy 3.1 is not released)

adrianeboyd:
I have a commit that switches the `spans_key` to `sc` to `evaluate` works. Would that be okay?

svlandeg:
Does `evaluate` only work with the default key? :(

adrianeboyd:
I'm sure we can improve it, but I didn't want to spend more time on it right now.

adrianeboyd:
What about having a custom evaluate script instead? If you call the underlying function in `evaluate` directly there is a parameter that should work now.

PR #28, opened 2021-01-21
Update REL project

svlandeg:
Fix REL project to latest spaCy & add unit test.

~~Unfortunately the REL project is broken in `v3.0.0rc3`. It requires the fix from https://github.com/explosion/spaCy/pull/6765 and should work again in spaCy `v3.0.0rc4`~~

~~Also fixes the goemotions project to use the new `textcat_multilabel` component (not necessary anymore because of https://github.com/explosion/projects/pull/30)~~


OpenAIMP:
Hi, can you please provide the command that is used to generate the config files for rel_component? https://github.com/explosion/projects/tree/v3/tutorials/rel_component/configs

PR #29, opened 2021-01-22
Update speed benchmarking project

svlandeg:
Three little fixes:
- Ensure the script continues if one library crashes (but still show the full stack trace)
- Add `token_splitter` component to spaCy's pretrained `en_core_web_trf` (temporary hack until we fix this more properly in the pipeline for the final v3)
- Limit batch size for the various processors of Stanza, which otherwise eats up 16GB of GPU memory and then crashes...

On CPU, Stanza does run without the batch size limitation. Comparing the eval on CPU with and without the batch size, it does seem to slightly suffer, from 1070 WPS (without limits) to 880 WPS (with limits) but overall it seems to be in the same ball park.


svlandeg:
Let's not merge yet - I want to double check things with the latest RC first

PR #30, opened 2021-01-29
fix textcat_goemotions textcat factory

timothyjlaurent:
As I was testing this out, I realized that it was using the incorrect factory for this multilabel dataset.


svlandeg:
Thanks for the PR! This is indeed a recent change in spaCy. Merging :-)

PR #34, opened 2021-02-02
fix: Fix text_cat_goemotions project init-vectors

reneoctavio:
Hello folks!

I was testing the text_cat_goemotions project. And the init-vectors command was not working.
It was throwing the following message: `Error: No such command 'model'.`

I noticed that the command changed (https://spacy.io/api/cli#init-vectors),
so I modified it accordingly. Seems to be working now.


svlandeg (approved):
Nice catch! This command is not run as part of the "main" workflow which is why we missed it in the tests. Thanks for the fix!

PR #38, opened 2021-02-17
Demo basic NER project

adrianeboyd:


adrianeboyd:
This will currently fail without https://github.com/explosion/spaCy/pull/7215.

PR #39, opened 2021-02-18
Demo textcat projects

adrianeboyd:
textcat:

* Data is textcat_docs_issues data in a simpler JSONL format

textcat_multilabel:

* Data is from cooking.stackexchange.com


svlandeg:
Would it be nice for the `multi_label` data to actually include examples with multiple true labels?

adrianeboyd:
Maybe in a separate project? I kind of like showing how these are basically equivalent, but maybe that's not particularly helpful in this kind of demo.

svlandeg:
I see your point, but I worry that it might actually be more confusing as it is now, because the data for the "single" classifier has both labels defined in each example, and the data for "multi" only has one label per gold instance...

adrianeboyd:
We can remove it here and create a separate multilabel project?

svlandeg:
Ah yea, maybe that's better. That also avoid the messing about in the project yaml (which also means the tests don't run for the multilabel)

PR #40, opened 2021-02-19
Update NEL prodigy script

svlandeg:
- NEL Prodigy annotation upgrade to spaCy v3 / Prodigy nightly
- remove redundant `init-config` from `ner_wikiner`


svlandeg:
Tests will fail until the next version of spaCy is released with [this fix](https://github.com/explosion/spaCy/pull/7074).

PR #41, opened 2021-03-01
Change command name to match docs

polm:
Docs refer to `preprocess` but it was called `convert` here.

Also removed a stale comment.

### DIFF
--- a/pipelines/tagger_parser_ud/project.yml
+++ b/pipelines/tagger_parser_ud/project.yml
@@ -26,16 +26,14 @@ assets:
 
 workflows:
   all:
-    - convert
+    - preprocess
     - train
     - evaluate
     - package
 
 commands:
-  - name: convert
+  - name: preprocess
     help: "Convert the data to spaCy's format"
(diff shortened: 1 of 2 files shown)


pipelines/tagger_parser_ud/project.yml
```diff
@@ -26,16 +26,14 @@ assets:
 
 workflows:
   all:
-    - convert
+    - preprocess
```
svlandeg, line 29: Good to have this sync'ed better with the spaCy docs, indeed.

Could you update the readme file as well? If you run the `update_docs.py` script from within the `.github` folder, it will automatically update all readme files. It might catch a few other outdated ones as well, that's fine ;-)

polm: Ah, got it - when I heard the script was in `.github` I thought it ran automatically :sweat_smile:

svlandeg: We probably could ;-) Then again I think there could be cases where you need to double check / adjust manually as well... Anyway, thanks! Merging :-)

PR #44, opened 2021-03-05
WIP: Fix benchmark/speed requirements in windows CI

adrianeboyd:


adrianeboyd:
So this is not exactly ideal, but I have no idea why the windows CI fails when downgrading numpy (due to flair in the speed benchmark project). This doesn't change any project requirements, just the numpy installed in the CI.

PR #50, opened 2021-03-22
Change testb.spacy to test.spacy so ner benchmark works

dogatekin:
In the CoNLL-03 named entity recognition benchmark, the test corpus is saved as `test.spacy` but it's later called as `testb.spacy` during the evaluation, which makes it fail. This PR fixes this issue.


svlandeg:
Good catch, thanks!

PR #57, opened 2021-04-30
Allow data in tsv with no labels

MatthiasMurray:
Allows custom .tsv datasets with some unlabeled data points to be processed by read_tsv function. If there is a text classification dataset where for one or more data points, no labels apply (and no dummy label has been generated for N/A), this allows those to be processed using the spacy projects run preprocess command.


svlandeg:
Thanks for the contribution!

PR #59, opened 2021-06-19
Add first take at double NER project

polm:
This project shows how to combine multiple trained NER pipelines.

Some unresolved things:

- how to use replace_listeners in code?
- should overwrite pipeline order on cli instead of using two configs


tutorials/ner_double/configs/drugs_second.cfg
```diff
@@ -0,0 +1,58 @@
...
+[nlp]
+lang = "en"
+# Note this is the only line that's different from the other config!
+pipeline = ["tok2vec","ner","ner_drug"]
+batch_size = 1000
+
+[components]
+
+[components.tok2vec]
+source = "en_core_web_md"
```
adrianeboyd, line 17: You don't need this, the `ner` has an internal `tok2vec`.

tutorials/ner_double/configs/drugs_first.cfg
```diff
@@ -0,0 +1,59 @@
...
+
+[nlp]
+lang = "en"
+pipeline = ["tok2vec","ner_drug","ner"]
+batch_size = 1000
+
+[components]
+
+[components.tok2vec]
+source = "en_core_web_md"
```
adrianeboyd, line 16: Ditto.

adrianeboyd:
https://spacy.io/api/language#replace_listeners

adrianeboyd:
Moving components should be easier, though. It's currently such a pain.

tutorials/ner_double/configs/drugs_second.cfg
polm: Ah, of course!

polm:
OK, the code works now, though it still gives a W113 error - is that normal or should I be doing things differently? 

Besides that I think this is ready to go, though I'd like someone to review the text (bottom of the README).

tutorials/ner_double/project.yml
```diff
@@ -0,0 +1,50 @@
...
+  Note that before running this project, you should run the
+  [ner_drugs](../ner_drugs) project and install the packaged version of it.
+
+vars:
+  name: "ner_double"
+  version: "0.0.0"
+  drug_model: "ner_drugs-0.0.0"
+
+
+# TODO is there a better name than "training" for assembled packages?
```
svlandeg, line 15: Maybe "pipelines"?

tutorials/ner_double/README.md
```diff
@@ -0,0 +1,114 @@
...
+
+The following commands are defined by the project. They
+can be executed using [`spacy project run [name]`](https://spacy.io/api/cli#project-run).
+Commands are only re-run if their inputs have changed.
+
+| Command | Description |
+| --- | --- |
+| `assemble` | Build two versions of the combined model from configs. |
+| `package` | Package the models so they can be installed. |
+| `visualize-model` | Use the model interactively using Streamlit |
```
svlandeg, line 25: I think in v3 we'd use the term "pipeline" here everywhere instead of "model". We've kind of changed that terminology in comparison to v2, as now the ML models in the pipeline components are more explicitly defined in the constructor, and things got confusing ;-)

polm: Makes perfect sense.

tutorials/ner_double/project.yml
```diff
@@ -0,0 +1,50 @@
...
+directories: ["configs", "scripts", "training", "packages"]
+
+workflows:
+  all:
+    - assemble
```
svlandeg, line 20: To be complete, perhaps this `project.yml` should also define the `download` command for the pretrained spacy pipeline, and define `en_ner_drugs` as an asset? Or where are you supposed to get the latter from? (maybe I missed it)

tutorials/ner_double/README.md
```diff
@@ -68,17 +68,22 @@ code. In code it looks like this:
 import spacy
 
 nlp = spacy.load("en_core_web_md") # load the base model
-drug = spacy.load("en_ner_drugs") # load the drug model
+drug_nlp = spacy.load("en_ner_drugs") # load the drug model
+# give this component a copy of its own tok2vec
+drug_nlp.replace_listeners("tok2vec", "ner", ["model.tok2vec"])
 
 # now you can put the drug model before or after the other ner
+# This will print a W113 warning but it's OK because we replaced listeners
```
adrianeboyd, line 76: The W113 warning isn't related to the listeners. It's related to the fact that `en_ner_drugs` doesn't have vectors in its vocab and `en_core_web_md` does. But as long as `en_ner_drugs` `ner` doesn't require vectors (`include_static_vectors = false`), it will work because it will just ignore the `en_core_web_md` vectors.

The problem is that there's no way to inspect whether a component references the vocab vectors, so all we can do is warn.

tutorials/ner_double/configs/double.cfg
```diff
@@ -0,0 +1,56 @@
...
+tolerance = 0.2
+
+[training.batcher.size]
+@schedules = "compounding.v1"
+start = 100
+stop = 1000
+compound = 1.001
+
+[initialize]
+vectors = null
```
adrianeboyd, line 56: Ah, this will break silently after https://github.com/explosion/spaCy/pull/8388.
```suggestion
vectors = "en_core_web_md"
```

adrianeboyd: Well, not silently. You'll get a warning that you **shouldn't** ignore and you'll get bad predictions (silently).

tutorials/ner_double/project.yml
polm: That's great, I'll use it!

tutorials/ner_double/project.yml
polm: My expectation was that a user would check out this repo, build the drug model separately and install it, and then use this tutorial. That's stated at the top of the README, but is there some way I can automate it in the `project.yml`? I guess I could make it cd to the other dir and call `spacy project` there but that feels weird...

polm:
Some of the tests failed here but it looks like there was a connection error during the speed test, so not an actual problem.

tutorials/ner_double/project.yml
svlandeg: No I wouldn't do that, but I was thinking maybe we could provide a prebuilt drug model somewhere that is just downloadable as an asset?

svlandeg:
I've restarted the CI ;-)

tutorials/ner_double/project.yml
polm: Ah, makes sense.

polm: I made a dummy release and added the drug model as an attachment, so it's an asset now. Much cleaner!

polm:
Once the tests go green this should be good to go.

PR #62, opened 2021-06-30
Update textcat_docs_issues

adrianeboyd:
* Switch to `textcat_multilabel`
* Switch to `spacy.TextCatEnsemble.v2`
* Remove train corpus `max_length`
* Add GPU option


svlandeg (approved):
Looks good to me. Perhaps a note in the readme somewhere why we're using `textcat_multilabel`, even if there are not multiple labels? 😁

PR #63, opened 2021-06-30
Add integration to Hugging Face Hub

osanseviero:
This PR adds an example project template with a `push_to_hub` script that uploads a packaged pipeline to the Hugging Face Hub. Using incoming spaCy release 3.1, it uses the generated `README.md` plus the `meta.json` to create the model card and use the right tags (type of pipeline, license, language). 

Here is an example repo (widget is not enabled yet)
https://huggingface.co/osanseviero/en_ner_fashion


osanseviero:
FYI @julien-c, @LysandreJik and @ines

integrations/huggingface_hub/scripts/push_to_hub.py
```diff
@@ -0,0 +1,106 @@
...
+
+def list_tags(name, tags):
+    if tags is None:
+        return ""
+    if isinstance(tags, str):
+        tags = [tags]
+    if len(tags) == 0:
+        return ""
+    name_tags = [f"- {tag}" for tag in tags]
+    return f"{name}:\n" + "\n".join(name_tags) + "\n"
```
julien-c, line 21: in `transformers` we switched to generating this via `pyyaml` which adds a dependency but is probably less error prone in the long term

integrations/huggingface_hub/scripts/push_to_hub.py
```diff
@@ -0,0 +1,106 @@
...
+        for component in data["components"]:
+            if component in token_classification_components and "token-classification" not in tags:
+                tags.append("token-classification")
+            if component in text_classification_components and "text-classification" not in tags:
+                tags.append("text-classification")
+
+        metadata = list_tags("tags", tags)
+        metadata += list_tags("language", lang)
+        metadata += list_tags("license", lic)
+        metadata_section = f"---\n{metadata}---\n"
```
julien-c, line 38: would be super neat if you're able to extract the metrics from the model card and also store them in the `model-index` format (https://github.com/huggingface/huggingface_hub/blame/main/modelcard.md) which will let the hub filter models by a metric e.g. NER F1 on a given dataset, etc.

integrations/huggingface_hub/scripts/push_to_hub.py
```diff
@@ -0,0 +1,106 @@
...
+
+    # Create model card, including HF tags
+    create_model_card(repo_local_path)
+
+    # Remove version from whl filename and rename to repo name
+    src_dir = os.path.join(package_path,f"{versioned_name}-py3-none-any.whl")
+    dst_file = os.path.join(repo_local_path, f"{lang_and_name}-any-py3-none-any.whl")
+    shutil.copyfile(src_dir, dst_file)
+
+    repo.push_to_hub(commit_message="Spacy Update")
```
julien-c, line 99: love this whole method's flow!

julien-c (commented):
this looks super cool! Love the integration with the reproducibility features from spaCy!

integrations/huggingface_hub/scripts/push_to_hub.py
osanseviero: Changed!

integrations/huggingface_hub/scripts/push_to_hub.py
osanseviero: I added support for the high level P/R/F metrics for NER, POS, dependencies and sentence segmentation. There are likely some metrics missing, I will double check.

It might also make sense to add the specific type metrics, but there is a lot of data there (see https://huggingface.co/spacy/en_core_web_sm/blob/main/en_core_web_sm-3.0.0/meta.json)

integrations/huggingface_hub/scripts/push_to_hub.py
```diff
@@ -0,0 +1,179 @@
...
+
+    model_index["results"] = results
+    return model_index
+
+
+def create_model_card(repo_name, repo_dir):
+    with open(os.path.join(repo_dir, "meta.json")) as f:
+        data = json.load(f)
+        lang = data["lang"] if data["lang"] != "xx" else "multilingual"
+        lic = data["license"].replace(" ", "-")
```
ines, line 96: `license` isn't technically a required field, so this is safer:

```suggestion
        lic = data.get("license", "").replace(" ", "-")
```

integrations/huggingface_hub/scripts/push_to_hub.py
```diff
@@ -0,0 +1,179 @@
...
+        metadata["model-index"]  = create_model_index(repo_name, data["performance"])
+        metadata = yaml.dump(metadata, sort_keys=False)
+
+
+        metadata_section = f"---\n{metadata}---\n"
+
+        # Read README generated by package
+        readme_path = os.path.join(repo_dir, "README.md")
+        readme = ""
+        with open(readme_path) as f:
```
ines, line 116: We might as well do a quick `exists` check here in case a user removed their `README.md` (they shouldn't, but I guess they should be allowed to 😛)

integrations/huggingface_hub/scripts/push_to_hub.py
osanseviero: What I saw is that when it's not there we get `"license":""`, so I think this should be fine. Committing for extra safety :)

integrations/huggingface_hub/scripts/push_to_hub.py
osanseviero: Done!

osanseviero:
@ines I did some changes to only consume the `.whl` instead of both the `.whl` and the `.tar`. This way, the user only specified the `.whl` and the repo is created. I also find the new repo structure clearer (`meta.json` and `config.cfg` are now at the top. There is also less duplicated info.

PR #72, opened 2021-08-02
Add wheel to azure-pipelines.yml

thomashacker:
## Issue
Projects that include `package` with `--build-wheel` in their workflow, fail the azure tests because `wheel` is not installed

## Change/Fix
- Add `wheel` as dependency in the `azure-pipeline.yml`


svlandeg:
Looks like the CI is unhappy unrelated, about the spancat project, so let's get this merged anyway.

PR #74, opened 2021-08-18
Change train-search in wandb to use Sweeps

scottire:
This PR adds [Weights & Biases Sweeps](https://docs.wandb.ai/guides/sweeps) as a replacement for the current `train-search` command implementation in `integrations/wandb`.

I've removed the existing scripts which generate multiple spaCy configs and run each pipeline. They are replaced with two scripts which can either run Sweeps using a sweep YAML config file or via the `wandb` python API. There's also a small bit of spaCy + thinc code which merges the spaCy + wandb sweeps configs, I'm sure there's other ways to do this, open to suggestions here. 

I've leaned towards using one ground truth spaCy config for the entire pipeline, with only the hyperparameters in question being in the sweeps config.


ines:
> I'm sure there's other ways to do this, open to suggestions here.

Not sure if this is a better approach, but `spacy train` also supports reading the `config.cfg` from standard input if you set it to `-`. So you can do something like:

```bash
python generate_config.py | spacy train -
```

All the `generate_config.py` script needs to do is output the generated config's string representation, e.g. `print(config.to_str())`. You could even take this one step further and make the script a CLI with `typer` so it could take different arguments/options/flags.

scottire:
Thanks for the suggestion. That's a really nice feature, I think using `spacy train` directly is great and most users probably want that. My only concern is that piping via stdout into `spacy train -` feels slightly magical and _less_ typical than just using the `train` function. I think a compromise would be to make this possible via a command line argument but default to use the `train` function. Would this suffice or what do you think?

PR #76, opened 2021-09-05
Initial version of polar embeddings

polm:
Couple of issues:

This is a stateful component, but it doesn't train. So maybe it
shouldn't be a tutorial project, but somewhere else?

I used the IMDB reviews dataset from Kaggle for this but it looks like
there's no public URL to download the same version of the data. The
upstream URL provided at Kaggle has the same data, but every tiny review
gets its own file so the overhead takes it from 60MB to 480MB on disk.
The dataset is too big anyway, so maybe there's a better alternative?

To calculate polar axes properly, the reference words should be averaged
with the nearest words in the vocab, but I'm not doing that yet.


svlandeg (commented):
I definitely like the idea of having more self-contained examples here!

> This is a stateful component, but it doesn't train. So maybe it
> shouldn't be a tutorial project, but somewhere else?

Maybe it's slightly more appropriate under `pipelines`? The distinction isn't always clear-cut though, and I don't really mind either way.

> I used the IMDB reviews dataset from Kaggle for this but it looks like
> there's no public URL to download the same version of the data. The
> upstream URL provided at Kaggle has the same data, but every tiny review
> gets its own file so the overhead takes it from 60MB to 480MB on disk.
> The dataset is too big anyway, so maybe there's a better alternative?

Either way it would be good to clarify to users how to get the data. You can have a readme file automatically generated & further edit that: https://github.com/explosion/projects#%EF%B8%8Frepository-maintanance (also note that the root dir readme should be updated - we might not have been doing this consistently). 

The failing unit tests are definitely unrelated to this PR, and something that's being fixed in spaCy, so no worries there.

tutorials/polar_component/scripts/polar_pipe.py
```diff
@@ -0,0 +1,106 @@
...
+
+    def get_key(self, sep="-"):
+        return f"{self.neg}{sep}{self.pos}"
+
+
+@Language.factory(
+    "polar",
+    requires=["doc.vector"],
+    default_config={},
+    default_score_weights={},  # ???
```
svlandeg, line 30: Why the comment here?

tutorials/polar_component/scripts/polar_pipe.py
```diff
@@ -0,0 +1,106 @@
...
+    return PolarEmbeddings(
+        nlp,
+        name,
+    )
+
+
+class PolarEmbeddings(Pipe):
+    """PolarEmbeddings let you turn normal word embeddings into embeddings
+    oriented along axes of meaning, preserving the overall distance of the
+    original embeddings while given dimensions semantic meaning.
```
svlandeg, line 45:
```suggestion
    original embeddings while giving dimensions semantic meaning.
```

tutorials/polar_component/scripts/evaluate.py
```diff
@@ -0,0 +1,68 @@
...
+
+    # now add some axes
+    polar.add_axis("science", "magic")
+    polar.add_axis("serious", "funny")
+    polar.add_axis("heartwarming", "creepy")
+    polar.add_axis("bad", "good")
+    return nlp
+
+def format_review(text, truncate=3):
+    #text = "    " + text # indent
```
svlandeg, line 24:
```suggestion
```

tutorials/polar_component/scripts/polar_pipe.py
polm: Actually I forget... I guess it can just be removed.

svlandeg:
Looks good to me! Feel free to merge when you're ready. 
Ah maybe one thing is to update the readme of the `pipelines` root dir (https://github.com/explosion/projects/tree/v3/pipelines) - there's another maintenance script for that :-)

PR #81, opened 2021-10-27
Add floret vectors demo

adrianeboyd:


adrianeboyd:
We can add tests here after v3.2.0 is released.

PR #91, opened 2021-12-17
Remove `entity` from `wandb.sweep` call

ash0ts:
This caused a `CommError` when run as is. Most likely caused by using an entity that did not match the API key provided which is the users own.


ash0ts:
This relates to wandb/examples#136

svlandeg:
Thanks!

PR #92, opened 2021-12-28
Update all configs to use Tok2Vec.v2 instead of v1

polm:
Not tested, though I don't think this should break anything.

We may want to do this for some other old architectures too, but for now
I just focused on this one.

Inspired by https://github.com/explosion/spaCy/discussions/9815.


svlandeg:
> Not tested, though I don't think this should break anything.

Famous last words ;-) 

Could you look into the failing tests? I'm guessing that `MaxoutWindowEncoder` should be bumped together with `Tok2Vec` - cf the original PR https://github.com/explosion/spaCy/pull/6442

PR #95, opened 2022-01-21
Fix directory creation problem in WandB example

ljvmiranda921:
Ref: https://github.com/explosion/spaCy/discussions/10097
Additionally, I formatted the affected code with black


svlandeg (approved):
Thanks, looks good to me!

thomashacker (approved):
Looks good! Thanks LJ 🚀

PR #99, opened 2022-03-14
Add floret Wikipedia+OSCAR vectors project

adrianeboyd:
Add a user-friendly cross-platform floret vectors project for training on Wikipedia (using `wikiextractor`) + OSCAR (using `datasets`).

The project defaults use Macedonian to provide a small but realistic demo.

The tests override the defaults to use the much-smaller Yoruba data sources.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -38,7 +38,8 @@ jobs:
                 pip install "spacy_lookups_data>=1.0.2,<1.1.0"
                 pip install torch==1.7.1+cpu -f https://download.pytorch.org/whl/torch_stable.html
                 # for pipelines/floret_wiki_oscar_vectors
-                pip install floret more-itertools datasets wikiextractor
+                pip install floret more-itertools datasets
+                pip install "wikiextractor @ git+https://github.com/adrianeboyd/wikiextractor.git@v3.0.7a0"
             displayName: 'Install dependencies'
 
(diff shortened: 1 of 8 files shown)


adrianeboyd:
This will require thinc v8.0.15 with https://github.com/explosion/thinc/pull/610.

MariamRiaz:
Where can we raise an issue about the projects ?? I am using fastapi to tag my data and I have a very large data set and when the limit reaches to around a million the fastapi stops upon more investigation it is a memory leak issue and the memory consumption is increasing as the dataset it processes increases.

adrianeboyd:
@MariamRiaz The [spacy discussions board](https://github.com/explosion/spacy/discussions) is the best place to start.

azure-pipelines.yml
```diff
@@ -31,12 +31,15 @@ jobs:
                 architecture: 'x64'
 
           - script: |
-                pip install "spacy>=3.1.0,<3.2.0"
+                pip install "spacy>=3.2.0,<3.3.0"
```
svlandeg, line 34: Do you think there will be something breaking in 3.3? Considering that release is near, it might be nice to make sure it's supported too?

adrianeboyd: Sure, but maybe in a separate PR after v3.3 is released?

PR #101, opened 2022-03-31
Add coref project

polm:
This adds a training config for the new coref component using OntoNotes / CoNLL 2012 data. 

The current config works, but is relatively brittle in relation to the OntoNotes setup. It also only trains the word-level coref component since the span resolver is not quite ready.

Hyperparameters are decent but have not been checked extensively.

Some things that should be done before this is merged:

- [x] better downloading for CoNLL 2012 data (doesn't work due to weird server settings)
- [x] verify python2 installation and OntoNotes paths in code
- [x] separate config with span predictor
-  ~~(maybe) support for gold span2head conversion like wl-coref (requires Java)~~ (this isn't needed)

Also, I put this in "experimental", but maybe it belongs somewhere else - I wasn't very sure about that.


svlandeg:
>Also, I put this in "experimental", but maybe it belongs somewhere else - I wasn't very sure about that.

`pipelines` ?

polm:
OK, I believe the issues with this project file specifically have been sorted out. 

The issues with the old coref scripts have been resolved by putting them in [their own repo](https://github.com/explosion/conll-2012). Some of the path-related issues were not inherent and were just a config issue.

I'm going to leave this in draft until the span predictor is wrapped up, but if you can install the `feature/coref` branch of spaCy then this should work.

polm:
At this point, building a full pipeline with the `feature/coref` branch of spaCy works, and should be possible for anyone with OntoNotes.

svlandeg (commented):
Thanks for all this, excited to play with it!

As I go through things, I'll be leaving reviewing comments in batches :-)

pipelines/coref/project.yml
```diff
@@ -0,0 +1,127 @@
...
+    - prep-span-data
+    - train-span-predictor
+    - assemble
+
+commands:
+  - name: "prep-data"
+    help: "Rehydrate the data using OntoNotes"
+    script: 
+      # XXX python2 is required for the old scripts
+      - python2 -c "print('Checking for Python 2...')"
```
svlandeg, line 53: This can hopefully be removed with https://github.com/explosion/conll-2012/pull/1

pipelines/coref/project.yml
```diff
@@ -0,0 +1,127 @@
...
+      #- tar xvzf assets/conll2012-coref/reference-coreference-scorers.v8.01.tar.gz -C assets/
+
+      - echo "Rehydrating data... This will take roughly 20 minutes"
+      - "bash assets/conll-2012/v3/scripts/skeleton2conll.sh -D ${vars.ontonotes} assets/conll-2012"
+
+      # Concatenate the data into single files to make it easier to work with
+      - sh -c "rm -f assets/*._gold_conll"
+      - sh -c "cat assets/conll-2012/v4/data/development/data/english/annotations/*/*/*/*.v4_gold_conll >> assets/dev.gold.conll"
+      - sh -c "cat assets/conll-2012/v4/data/train/data/english/annotations/*/*/*/*.v4_gold_conll >> assets/train.gold.conll"
+      - sh -c "cat assets/conll-2012/v4/data/test/data/english/annotations/*/*/*/*.v4_gold_conll >> assets/test.gold.conll"
```
svlandeg, line 67: These commands don't seem to work on Windows, because of an issue we have with quotes. I think this might also be the case on Mac, we have to check. The commands themselves are fine though.

Running through `spacy projects`: 

```
Running command: sh -c '"cat assets/conll-2012/v4/data/development/data/english/annotations/*/*/*/*.v4_gold_conll >> assets/dev.gold.conll"'
sh: cat assets/conll-2012/v4/data/development/data/english/annotations/*/*/*/*.v4_gold_conll >> assets/dev.gold.conll: No such file or directory
```

Running directly on console, no issue, and the 3 files are properly created:
```
$ sh -c "cat assets/conll-2012/v4/data/development/data/english/annotations/*/*/*/*.v4_gold_conll >> assets/dev.gold.c
$ sh -c "cat assets/conll-2012/v4/data/train/data/english/annotations/*/*/*/*.v4_gold_conll >> assets/train.gold.conll
$ sh -c "cat assets/conll-2012/v4/data/test/data/english/annotations/*/*/*/*.v4_gold_conll >> assets/test.gold.conll"
```

pipelines/coref/project.yml
```diff
@@ -0,0 +1,127 @@
...
+
+workflows:
+  prep:
+    - prep-data
+    - preprocess
+  train:
+    - train-cluster
+    - prep-span-data
+    - train-span-predictor
+    - assemble
```
svlandeg, line 36: This is really weird, but I ran `spacy project run train` today and it ran `train-cluster`, trained the model, saved it, and then quit. `prep-span-data` etc weren't executed?

```
(venv_spacy)
Sofie@DESKTOP-NATQ3P4 MINGW64 .../projects/pipelines/coref (feature/coref)
$ spacy project run train
ℹ Running workflow 'train'

=============================== train-cluster ===============================
Running command: '...python.exe' -m spacy train configs/cluster.cfg -g 0 --paths.train corpus/train.spacy --paths.dev corpus/dev.spacy -o training/cluster --training.max_epochs 5
ℹ Saving to output directory: training\cluster
ℹ Using GPU: 0

=========================== Initializing pipeline ===========================
[2022-04-29 11:58:25,012] [INFO] Set up nlp object from config
...
✔ Initialized pipeline

============================= Training pipeline =============================
ℹ Pipeline: ['sentencizer', 'transformer', 'coref']
ℹ Set annotations on update for: ['sentencizer']
ℹ Initial learn rate: 0.0001
E    #       LOSS TRANS...  LOSS COREF  SENTS_F  COREF_F  COREF_P  COREF_R  SCORE
---  ------  -------------  ----------  -------  -------  -------  -------  ------
  0       0           0.00      204.01    83.20     0.00     0.00     0.00    0.00
  0    1000           0.00    36042.83    83.20    35.34    52.41    26.65    0.35
...
  2    7000           0.00    22403.98    83.20    56.39    73.86    45.61    0.56
✔ Saved pipeline to output directory
training\cluster\model-last

(venv_spacy)
Sofie@DESKTOP-NATQ3P4 MINGW64 .../projects/pipelines/coref (feature/coref)
```

polm: That is really weird... Could it be Windows related? 

I assume the training must have exited with a non-zero code but I can't imagine why if the model was saved.

pipelines/coref/project.yml
polm: Thanks, removed it.

pipelines/coref/project.yml
```diff
@@ -0,0 +1,117 @@
+title: "Training a spaCy Coref Model"
+description: >
+  This project trains a coreference model for spaCy using OntoNotes.
+
+vars:
+  # XXX Change to your actual GPU ID
+  gpu_id: 0
+  # XXX fill with your local path to OntoNotes
+  ontonotes: "/home/USER/ontonotes5/data"
+  max_epochs: 20
```
svlandeg, line 10: Is it reasonable to assume that this will be the same value for `train_cluster` and `train-span-predictor`?

pipelines/coref/project.yml
```diff
@@ -0,0 +1,117 @@
...
+    script:
+      - spacy train configs/span.cfg -c scripts/custom_functions.py -g ${vars.gpu_id} --paths.train corpus/spans.train.spacy --paths.dev corpus/spans.dev.spacy --training.max_epochs ${vars.max_epochs} --paths.transformer_source training/cluster/model-best -o training/span_predictor
+    deps:
+      - corpus/spans.train.spacy
+      - corpus/spans.dev.spacy
+    outputs:
+      - training/span_predictor
+
+  - name: "assemble"
+    help: "Assemble parts into complete coref pipeline."
```
svlandeg, line 109:
```suggestion
    help: "Assemble all parts into a complete coref pipeline."
```

pipelines/coref/scripts/create_eval_data.py
```diff
@@ -0,0 +1,93 @@
+#!/usr/bin/env python3
+# Read in a file and produce evaluation data using a model.
+# Evaluation data will be a `.response` and `.key` file.
+
+import os
```
svlandeg, line 5:
```suggestion
```

pipelines/coref/scripts/create_eval_data.py
```diff
@@ -0,0 +1,93 @@
+#!/usr/bin/env python3
```
svlandeg, line 1: Is this script still being used? I don't see it referenced from the yml file? Similarly, do we still need this line in the yml?
```
rm -f corpus/test.key corpus/test.response
```

pipelines/coref/project.yml
```diff
@@ -0,0 +1,117 @@
...
+
+  all:
+    # prep
+    - prep-data
+    - preprocess
+    # train
+    - train-cluster
+    - prep-span-data
+    - train-span-predictor
+    - assemble
```
svlandeg, line 42:
```suggestion
    - assemble
    # eval
    - eval
```

pipelines/coref/configs/cluster.cfg
```diff
@@ -0,0 +1,136 @@
...
+init_tok2vec = null
+vectors = null
+
+[system]
+gpu_allocator = "pytorch"
+seed = 0
+
+[nlp]
+lang = "en"
+pipeline = ["sentencizer","transformer", "coref"]
```
svlandeg, line 13:
```suggestion
pipeline = ["sentencizer", "transformer", "coref"]
```

pipelines/coref/configs/coref.cfg
```diff
@@ -0,0 +1,34 @@
...
+init_tok2vec = null
+vectors = null
+
+[system]
+gpu_allocator = "pytorch"
+seed = 0
+
+[nlp]
+lang = "en"
+pipeline = ["sentencizer","transformer", "coref", "span_predictor"]
```
svlandeg, line 13:
```suggestion
pipeline = ["sentencizer", "transformer", "coref", "span_predictor"]
```

pipelines/coref/project.yml
```diff
@@ -0,0 +1,117 @@
...
+    help: "Prepare data for the span predictor component."
+    script:
+      - python scripts/reader.py --model-path training/cluster/model-best/ --gpu --input-path corpus/train.spacy --output-path corpus/spans.train.spacy --head-prefix coref_head_clusters --span-prefix coref_clusters
+      - python scripts/reader.py --model-path training/cluster/model-best/ --gpu --input-path corpus/dev.spacy --output-path corpus/spans.dev.spacy --head-prefix coref_head_clusters --span-prefix coref_clusters
+    deps:
+      - corpus/train.spacy
+      - corpus/dev.spacy
+    outputs:
+      - corpus/spans.train.spacy
+      - corpus/spans.dev.spacy
```
svlandeg, line 96: For readability, I'd pick one format for `deps` and `outputs` throughout the yml file (with or without quotes around the paths consistently)

pipelines/coref/project.yml
```diff
@@ -0,0 +1,117 @@
...
+  - name: "train-cluster"
+    help: "Train the clustering component"
+    script: 
+      - "python -m spacy train configs/cluster.cfg -g ${vars.gpu_id} --paths.train corpus/train.spacy --paths.dev corpus/dev.spacy -o training/cluster --training.max_epochs ${vars.max_epochs}"
+    deps:
+      - "corpus/train.spacy"
+      - "corpus/dev.spacy"
+      - "configs/cluster.cfg"
+    outputs:
+      - "training/model-best"
```
svlandeg, line 84:
```suggestion
      - "training/cluster/model-best"
```

pipelines/coref/project.yml
```diff
@@ -0,0 +1,117 @@
...
+  
+  - name: "prep-span-data"
+    help: "Prepare data for the span predictor component."
+    script:
+      - python scripts/reader.py --model-path training/cluster/model-best/ --gpu --input-path corpus/train.spacy --output-path corpus/spans.train.spacy --head-prefix coref_head_clusters --span-prefix coref_clusters
+      - python scripts/reader.py --model-path training/cluster/model-best/ --gpu --input-path corpus/dev.spacy --output-path corpus/spans.dev.spacy --head-prefix coref_head_clusters --span-prefix coref_clusters
+    deps:
+      - corpus/train.spacy
+      - corpus/dev.spacy
```
svlandeg, line 93:
```suggestion
      - corpus/dev.spacy
      - training/cluster/model-best
```

pipelines/coref/project.yml
```diff
@@ -0,0 +1,117 @@
...
+      - corpus/spans.train.spacy
+      - corpus/spans.dev.spacy
+  
+  - name: "train-span-predictor"
+    help: "Train the span predictor component."
+    script:
+      - spacy train configs/span.cfg -c scripts/custom_functions.py -g ${vars.gpu_id} --paths.train corpus/spans.train.spacy --paths.dev corpus/spans.dev.spacy --training.max_epochs ${vars.max_epochs} --paths.transformer_source training/cluster/model-best -o training/span_predictor
+    deps:
+      - corpus/spans.train.spacy
+      - corpus/spans.dev.spacy
```
svlandeg, line 104:
```suggestion
      - corpus/spans.dev.spacy
      - training/cluster/model-best
```

pipelines/coref/project.yml
```diff
@@ -0,0 +1,117 @@
...
+    outputs:
+      - training/span_predictor
+
+  - name: "assemble"
+    help: "Assemble parts into complete coref pipeline."
+    script:
+      - spacy assemble configs/coref.cfg training/coref
```
svlandeg, line 111:
```suggestion
      - spacy assemble configs/coref.cfg training/coref
    deps:
      - training/cluster/model-best
      - training/span_predictor/model-best
```

svlandeg (commented):
Some smaller comments about the project workflow

pipelines/coref/project.yml
```diff
@@ -0,0 +1,117 @@
...
+  
+  - name: "train-span-predictor"
+    help: "Train the span predictor component."
+    script:
+      - spacy train configs/span.cfg -c scripts/custom_functions.py -g ${vars.gpu_id} --paths.train corpus/spans.train.spacy --paths.dev corpus/spans.dev.spacy --training.max_epochs ${vars.max_epochs} --paths.transformer_source training/cluster/model-best -o training/span_predictor
+    deps:
+      - corpus/spans.train.spacy
+      - corpus/spans.dev.spacy
+    outputs:
+      - training/span_predictor
```
svlandeg, line 106:
```suggestion
      - training/span_predictor/model-best
```

pipelines/coref/project.yml
polm: Not sure exactly which one you're referring to, but yes for the gpu and ontonotes path. 

For the epochs, currently it makes sense to keep them the same. In practice the span predictor trains much more quickly, so the limit could be lower there, but because of the patience setting it doesn't really matter that the max epochs are high.

pipelines/coref/scripts/create_eval_data.py
polm: You're right, I think that line and this script are no longer necessary. I'm kind of tempted to leave the script around in case someone wanted to run the official eval scripts on their own, but I'll remove it for now.

pipelines/coref/project.yml
svlandeg: Ah yes I meant the `max_epochs` only, I guess Github just shows a few previous lines by default. Ok thanks!

polm:
Test failures were due to speed tests on other components, I expect they will be resolved by merge. 

To clarify how to run this PR, you should install vanilla spaCy (from PyPI or a local dev env / master) and [the branch in this spacy-experimental PR](https://github.com/explosion/spacy-experimental/pull/17). After that you need to edit the config to include a path to your local copy of OntoNotes. Everything else should be handled by the project file.

adrianeboyd:
I tried this out starting in a clean venv. This project will need `requirements.txt` that includes the right version of `spacy-experimental` (you could just initially point to the right `git+https` or `archive` URL?) and also to specify that you need `en_core_web_sm` somehow.

The first error I ran into:

```
================================= preprocess =================================
Running command: /tmp/venv38-1/bin/python3.8 scripts/preprocess.py assets/train.gold.conll corpus/train.spacy
Traceback (most recent call last):
  File "scripts/preprocess.py", line 109, in <module>
    read_file(sys.argv[1], sys.argv[2])
  File "scripts/preprocess.py", line 19, in read_file
    nlp = spacy.load("en_core_web_sm", disable=["tagger", "ner", "attribute_ruler", "lemmatizer"])
  File "/tmp/venv38-1/lib/python3.8/site-packages/spacy/__init__.py", line 51, in load
    return util.load_model(
  File "/tmp/venv38-1/lib/python3.8/site-packages/spacy/util.py", line 427, in load_model
    raise IOError(Errors.E050.format(name=name))
OSError: [E050] Can't find model 'en_core_web_sm'. It doesn't seem to be a Python package or a valid path to a data directory.
```

polm:
Thanks for the feedback! I think I have fixed most of what you pointed out. 

> Also maybe torch as a requirement somehow? And spacy-transformers.

I specified `spacy-transformers` as a requirement, and I think that pulls in `torch`, but I'm not sure how to specify the right GPU stuff - for example, GPU version will depend on the user's CUDA version. I can call nvcc or something but maybe it's easier to just say "please install spaCy with GPU support" and have a short script check for it and fail at the start?

adrianeboyd:
The spacy version should go in `spacy_version` in `project.yml` instead?

pipelines/coref/project.yml
```diff
@@ -0,0 +1,133 @@
...
+    help: "Convert the data to spaCy's format"
+    deps:
+      - assets/train.gold.conll
+      - assets/dev.gold.conll
+      - assets/test.gold.conll
+    script:
+      - python scripts/preprocess.py assets/train.gold.conll corpus/train.spacy
+      - python scripts/preprocess.py assets/dev.gold.conll corpus/dev.spacy
+      - python scripts/preprocess.py assets/test.gold.conll corpus/test.spacy
+    - outputs:
```
adrianeboyd, line 82:
```suggestion
    outputs:
```

adrianeboyd:
There's some renaming missing for `span_predictor` in terms of the configs and output directories (`assemble` fails).

If I rename enough to assemble, then:

```
==================================== eval ====================================
Running command: /tmp/venv38-1/bin/python3.8 scripts/run_eval.py training/coref corpus/test.spacy
Traceback (most recent call last):
  File "scripts/run_eval.py", line 7, in <module>
    from spacy.coref_scorer import Evaluator, get_cluster_info, lea
ModuleNotFoundError: No module named 'spacy.coref_scorer'
```

pipelines/coref/scripts/preprocess.py
```diff
@@ -0,0 +1,109 @@
...
+from spacy.tokens import Doc, DocBin
+
+DOCID_REGEX = "#begin document \((.*)\); part (\d*)"
+
+def read_file(fname, outname):
+    with open(fname) as infile:
+        text = infile.read()
+
+    # TODO use larger model
+    nlp = spacy.load("en_core_web_sm", disable=["tagger", "ner", "attribute_ruler", "lemmatizer"])
```
kadarakos, line 19:
```suggestion
    nlp = spacy.load("en_core_web_lg", disable=["tagger", "ner", "attribute_ruler", "lemmatizer"])
```

pipelines/coref/project.yml
```diff
@@ -0,0 +1,133 @@
...
+      - assets/dev.gold.conll
+      - assets/test.gold.conll
+
+  - name: "preprocess"
+    help: "Convert the data to spaCy's format"
+    deps:
+      - assets/train.gold.conll
+      - assets/dev.gold.conll
+      - assets/test.gold.conll
+    script:
```
kadarakos, line 78:
```suggestion
    script:
        - python -m spacy download en_core_web_lg
```

pipelines/coref/configs/cluster.cfg
```diff
@@ -0,0 +1,135 @@
...
+before_to_disk = null
+
+[training.batcher]
+@batchers = "spacy.batch_by_sequence.v1"
+size = 1
+get_length = null
+
+[training.logger]
+@loggers = "spacy.ConsoleLogger.v1"
+progress_bar = false
```
kadarakos, line 102: I kinda like having it on more, because I would like to know timing plus one epoch takes a long time and it seems a bit like its hanging on the first epoch -- to me at least.

pipelines/coref/configs/cluster.cfg
```diff
@@ -0,0 +1,135 @@
...
+beta2 = 0.999
+L2_is_weight_decay = true
+L2 = 0.01
+grad_clip = 1.0
+use_averages = true
+eps = 1e-8
+learn_rate = 0.0001
+
+[training.score_weights]
+sents_f = 0
```
kadarakos, line 116: Why do we care about the sentence f-score? We do not train a sentence boundary predictor right?

pipelines/coref/configs/cluster.cfg
```diff
@@ -0,0 +1,135 @@
...
+init_tok2vec = null
+vectors = null
+
+[system]
+gpu_allocator = "pytorch"
+seed = 0
+
+[nlp]
+lang = "en"
+pipeline = ["sentencizer", "transformer", "coref"]
```
kadarakos, line 13: I don't think we need the `sentencizer` for this component only for `span-resolver`.

pipelines/coref/configs/coref.cfg
```diff
@@ -0,0 +1,34 @@
...
+init_tok2vec = null
+vectors = null
+
+[system]
+gpu_allocator = "pytorch"
+seed = 0
+
+[nlp]
+lang = "en"
+pipeline = ["sentencizer", "transformer", "coref", "span_predictor"]
```
kadarakos, line 13:
```suggestion
pipeline = ["sentencizer", "transformer", "coref", "span_resolver"]
```

pipelines/coref/configs/coref.cfg
```diff
@@ -0,0 +1,34 @@
...
+after_pipeline_creation = null
+tokenizer = {"@tokenizers":"spacy.Tokenizer.v1"}
+
+[components]
+
+[components.sentencizer]
+factory = "sentencizer"
+punct_chars = null
+
+[components.span_predictor]
```
kadarakos, line 27:
```suggestion
[components.span_resolver]
```

pipelines/coref/configs/coref.cfg
```diff
@@ -0,0 +1,34 @@
...
+source = "training/span_predictor/model-best"
```
kadarakos, line 28:
```suggestion
source = "training/span_resolver/model-best"
```

pipelines/coref/project.yml
```diff
@@ -0,0 +1,133 @@
...
+      - corpus/dev.spacy
+      - training/cluster/model-best
+    outputs:
+      - corpus/spans.train.spacy
+      - corpus/spans.dev.spacy
+  
+  - name: "train-span-predictor"
+    help: "Train the span predictor component."
+    script:
+      - spacy train configs/span.cfg -c scripts/custom_functions.py -g ${vars.gpu_id} --paths.train corpus/spans.train.spacy --paths.dev corpus/spans.dev.spacy --training.max_epochs ${vars.max_epochs} --paths.transformer_source training/cluster/model-best -o training/span_predictor
```
kadarakos, line 114:
```suggestion
      - spacy train configs/span.cfg -c scripts/custom_functions.py -g ${vars.gpu_id} --paths.train corpus/spans.train.spacy --paths.dev corpus/spans.dev.spacy --training.max_epochs ${vars.max_epochs} --paths.transformer_source training/cluster/model-best -o training/span_resolver
```

pipelines/coref/project.yml
```diff
@@ -0,0 +1,133 @@
...
+    deps:
+      - corpus/spans.train.spacy
+      - corpus/spans.dev.spacy
+      - training/cluster/model-best
+    outputs:
+      - training/span_predictor/model-best
```
kadarakos, line 120:
```suggestion
      - training/span_resolver/model-best
```

pipelines/coref/project.yml
```diff
@@ -0,0 +1,133 @@
...
+
+  - name: "assemble"
+    help: "Assemble all parts into a complete coref pipeline."
+    script:
+      - spacy assemble configs/coref.cfg training/coref
+    deps:
+      - training/cluster/model-best
+      - training/span_predictor/model-best
```
kadarakos, line 128:
```suggestion
      - training/span_resolver/model-best
```

pipelines/coref/configs/cluster.cfg
polm: Coref only resolves within sentences, so if the sentence f-score is bad it can have a large effect on performance. It's not necessary though.

pipelines/coref/configs/cluster.cfg
polm: I think we usually keep this false for demo models? I don't like it because I resize my terminal when training and the progress bar becomes a mess.

pipelines/coref/configs/cluster.cfg
polm: That's correct, I'll remove it.

pipelines/coref/project.yml
polm: Currently we are specifying models in `requirements.txt`. Still not clear if that's the best way though.

polm:
Thanks for all the feedback! I had been focused on testing the training part, but it looks like there was a lot of cleanup left. 

I am doing a test run to confirm everything locally, but I think it should work at this point. 

The remaining thing is how to specify a Torch / GPU lib install - I'm not really sure what the right way to do that is.

polm:
My test run of training and evaluation with the above changes finished successfully.

pipelines/coref/project.yml
polm: This should be addressed [here](https://github.com/explosion/projects/pull/101/commits/a3814210283d363c80a45a9cd77e199b7dcf6ade).

pipelines/coref/project.yml
polm: Can you let me know if this still happens? I have been keeping an eye out, but I have never seen it happen and I am not sure what could cause it. If it does happen I guess I'd want to check the exit code of the process.

pipelines/coref/project.yml
polm: Marking resolved because this is using the large one.

pipelines/coref/project.yml
svlandeg: Trying this again on a fresh system today!

pipelines/coref/README.md
```diff
@@ -0,0 +1,54 @@
...
+### 🗂 Assets
+
+The following assets are defined by the project. They can
+be fetched by running [`spacy project assets`](https://spacy.io/api/cli#project-assets)
+in the project directory.
+
+| File | Source | Description |
+| --- | --- | --- |
+| `assets/` | Git | CoNLL-2012 scripts and dehydrated data. |
+| `/home/USER/ontonotes5/data` | Local | Ensure you have a local copy of OntoNotes: https://catalog.ldc.upenn.edu/LDC2013T19 |
```
svlandeg, line 52: I was copying this to a clean system and wondering: it doesn't work if you only store the English data, right? I was hoping it would so I didn't need to copy all of it, but `prep-data` errors with 
```
$ spacy project run prep-data
"Rehydrating data... This will take roughly 20 minutes"
----------------------------------------------------------------------------------------------------

could not find the gold parse [c://data/ontonotes5/data/arabic/annotations/nw/ann/00/ann_0001.parse] in the ontonotes distribution ... exiting ...

----------------------------------------------------------------------------------------------------
cat: 'assets/conll-2012/v4/data/development/data/english/annotations/*/*/*/*.v4_gold_conll': No such file or directory
```

Just wondering if maybe there's an easy fix to allow users to have only the English data, as I assume this project only  uses the English data, too.

svlandeg: (this is not a showstopper though!)

pipelines/coref/requirements.txt
```diff
@@ -0,0 +1,3 @@
+spacy-transformers>=1.1.6,<1.2.0
+https://github.com/explosion/spacy-models/releases/download/en_core_web_lg-3.3.0/en_core_web_lg-3.3.0-py3-none-any.whl
+git+https://github.com/explosion/spacy-experimental@feature/coref#egg=spacy-experimental
```
svlandeg, line 3: Does the readme tell users to install these requirements? As installing them is not part of a workflow?

svlandeg: Also I think this is not quite right. If you have a custom install of `spacy`, this will pull in the latest install from `pypi` instead :(

svlandeg: I think we should updat all of this to be compatible with 3.4 & the new models, as people trying out experimental features are likely to have the latest `spacy` version,  IMO

pipelines/coref/project.yml
svlandeg: Unfortunately I couldn't replicate or refute this on the fresh system as it ran out of CUDA memory during training (which was sort of expected, it's only a laptop) and I don't have access to my normal/old/proper desktop system for at least another few days.

pipelines/coref/README.md
polm: I had not checked that, thanks for noting it! We definitely only need the English data, so I'll see how hard it would be to modify the old scripts to deal with this.

pipelines/coref/requirements.txt
polm: I think there are several unresolved questions around how to handle this that are bigger than this PR. 

1. **Do we tell users to install this, or do we automate it?** I think it makes sense to either tell users to `pip install -r requirements.txt` or assume they will know to do it anyway. In some projects we had the project do this (or similar) for the user, but we can't automate everything because projects don't run without spaCy. #121 is related. In particular I think it makes sense to have the user be partly responsible for this because I don't think there's a good way to automate GPU library installation. 
2. **spaCy 3.3.** I was actually testing this on 3.4 but lowered it for this PR due to the version pin in `spacy-experimental`. The pin has just been raised in the PR there so I can raise it here as well.

> Also I think this is not quite right. If you have a custom install of spacy, this will pull in 3.3.1 from pypi instead :(

I did not realize that would happen, I'll need to see which part of the requirements is doing that. Or we can do something like just explicitly state in the README "please install spaCy with GPU support before running this".

svlandeg: While we're figuring this out for all the projects, maybe a quick note in the readme and/or other guidelines we'd distritbute would work? For 3.4 I see that the link to the models has been upgraded so that's fine :-)

pipelines/coref/project.yml
svlandeg: I suggest we keep this in the back of our mind though, in case we see reports of something similar once we release this. I'll resolve in the meantime.

pipelines/coref/requirements.txt
polm: I updated the README to add instructions for this. 

I'm still unclear how spaCy with GPU install should work. We can't specify that in requirements because it depends on the user's hardware. Maybe the steps should be 1. install spaCy with GPU 2. install requirements, and the requirements shouldn't include spaCy directly?

pipelines/coref/README.md
polm: I have not tested running the modified script yet, but I added an optional `sed` command to the `project.yml` that I'm pretty sure can be used to only do the English files.

pipelines/coref/requirements.txt
polm: Currently I have modified the README to be:

1. install spaCY with GPU support
2. install requirements
3. modify `project.yml` (for paths, GPU ID)

Let me know if there's a better way to do that.

pipelines/coref/README.md
polm: I verified this works and turned it on by default since it speeds up preprocessing.

pipelines/coref/requirements.txt
polm: I assume the current state of the README is OK, let me know if not.

polm:
Just a warning, but since the components are in experimental now, I'm going to move this from `pipelines` to `experimental`, which will be disruptive if you have a local copy.

experimental/coref/project.yml
```diff
@@ -0,0 +1,175 @@
...
+      # We only need the English data
+      - sed -i 's/arabic.english.chinese/english/' assets/conll-2012/v3/scripts/skeleton2conll.sh
+
+      - echo "Rehydrating data... This will take a few minutes"
+      - "bash assets/conll-2012/v3/scripts/skeleton2conll.sh -D \"${vars.ontonotes}\" assets/conll-2012"
+
+      # Concatenate the data into single files to make it easier to work with
+      - bash scripts/concat_data.sh
+    deps:
+      - ${vars.ontonotes}
```
svlandeg, line 104:
```suggestion
    deps:
      - assets/conll-2012-development.v4.tar.gz
      - assets/conll-2012-test-key.tar.gz
      - assets/conll-2012-test-official.v9.tar.gz
      - assets/conll-2012-train.v4.tar.gz
      - assets/conll-2012/v3/scripts/skeleton2conll.sh
      - ${vars.ontonotes}
```
By adding these `deps` in specifically we avoid the generic crash
```
$ spacy project run prep-data
================================= prep-data =================================
Running command: tar xzf assets/conll-2012-development.v4.tar.gz -C assets/
tar: Error opening archive: Failed to open 'assets/conll-2012-development.v4.tar.gz'
```
if you forgot to run `spacy project assets` and instead get
```
$ spacy project run prep-data

✘ Missing dependency specified by command 'prep-data':
assets/conll-2012-development.v4.tar.gz
Maybe you forgot to run the 'project assets' command or a previous step?
```

experimental/coref/project.yml
```diff
@@ -0,0 +1,175 @@
...
+      - tar xzf assets/conll-2012-development.v4.tar.gz -C assets/
+      - tar xzf assets/conll-2012-test-key.tar.gz -C assets/
+      - tar xzf assets/conll-2012-test-official.v9.tar.gz -C assets/
+      - tar xzf assets/conll-2012-train.v4.tar.gz -C assets/
+
+      # We only need the English data
+      - sed -i 's/arabic.english.chinese/english/' assets/conll-2012/v3/scripts/skeleton2conll.sh
+
+      - echo "Rehydrating data... This will take a few minutes"
+      - "bash assets/conll-2012/v3/scripts/skeleton2conll.sh -D \"${vars.ontonotes}\" assets/conll-2012"
```
svlandeg, line 99: This command doesn't work on my Windows box, it's still unhappy with the quotes. The following does work for me, but please double check on your system:
```suggestion
      - bash assets/conll-2012/v3/scripts/skeleton2conll.sh -D ${vars.ontonotes} assets/conll-2012
```
Is there any way we can somehow run this project on a very limited set of data in the CI? We need to make sure it works properly across the main OS & Python versions.

experimental/coref/project.yml
```diff
@@ -0,0 +1,175 @@
...
+
+      # Concatenate the data into single files to make it easier to work with
```
svlandeg, line 101:
```suggestion
      # Concatenate the data into single train/dev/test files to make it easier to work with
```

experimental/coref/project.yml
polm: Thanks, that is an improvement!

experimental/coref/project.yml
polm: Sorry I missed that, I thought I'd resolved that issue. 

First, I'm going to go ahead and move this block into a bash script so that quoting issues won't come up again. (Unfortunately I think `${vars.ontonotes}` needs to be in quotes, especially for Windows, in case the path has a space in it.)

For the testing, I was wondering how to do that. What I've done for now is set up a run using a single document from [LitBank](https://github.com/dbamman/litbank), which has has files in the CoNLL format, to test most steps of the pipeline. But for this particular section the scripts are OntoNotes specific, so I'd have to copy the OntoNotes directory structure... I'll see if I can think of a way to test it.

polm:
The test I just added is failing in CI - the logs aren't clear, but I think it's just an issue of requirements not being installed in the test env. I'll sort it out Monday.

pipelines/coref/project.yml
svlandeg: Ugh, unfortunately this happened again.
```
$ spacy project run train
(...)
  6   19000           0.00    16163.36    67.59    69.78    65.54    0.68
  7   20000           0.00    15567.39    67.79    71.76    64.24    0.68
  7   21000           0.00    16047.34    67.66    69.10    66.28    0.68
  7   22000           0.00    15536.11    67.47    69.54    65.52    0.67
✔ Saved pipeline to output directory
training\cluster\model-last
``` 
The script did save the models, but then just exited without running the remainders of the commands. I suspect there's some reason why it crashes, but this error gets masked by the project system somehow.

svlandeg: Ok, did a few more tests, and it looks like my hunch was correct: there was not enough space on my system to safe the very final model, but this error didn't make it to the console somehow. When I try again with smaller data/less iterations, it "trains" the model, then does continue to `prep-span-data`. So this is probably more of a UX issue with `spacy project` or maybe even with `spacy train` then it is an issue with this specific project.

polm:
The previous test failures were due to GPU-related issues. Those are resolved but this is still failing, and I think it's due to disk space related issues (like the strange error upthread). I'm still looking into how to resolve it.

polm:
After debugging, I'm still not 100% sure what exit code -9 means, but it looks like this might be an issue with memory and not disk usage. I've changed the tests to use configs with tok2vecs instead of Transformers to test that.

polm:
Test works now, so that should make sure this runs on Windows. 

One thing that came up during the test is that this repo uses env vars to set overrides on `training.max_epochs` and `training.max_iterations`. This is a problem when running `spacy project assemble`, since that command doesn't allow overriding those values and will result in an error. The fix in this case was just to overwrite that env var before running the test, but I expect this will come up in other projects eventually.

svlandeg:
Thanks for getting the test in place! So currently this is running on CPU with tok2vec, right? Is that simply because there's no GPU instance available on the CI?

polm:
That's correct, it's running with tok2vec on CPU. That's partly because there's no GPU instance, and partly to keep memory and disk usage down so that the job doesn't die unpredictably.

experimental/coref/project.yml
polm: The quoting issue here and earlier are related to this issue: https://github.com/explosion/spaCy/pull/11378

polm:
Going to close and re-open to pull in latest changes in experimental PR...

polm:
It looks like the tests are failing on Linux due to memory usage when training the SpanResolver. None of the recent changes should modify that so maybe it's just flaky, but I'll see if I can reduce the memory usage further.

experimental/coref/test_project_coref.py
```diff
@@ -0,0 +1,22 @@
...
+from spacy.cli.project.assets import project_assets
+from pathlib import Path
+
+
+def test_experimental_coref():
+
+    # This is set globally in testing to make sure tests are fast, but can't be
+    # used for this project because passing training-specific parameters breaks
+    # the spacy assemble command.
+    os.environ["SPACY_CONFIG_OVERRIDES"] = ""
```
svlandeg, line 12: But then any test run after this will not have this setting anymore?

experimental/coref/scripts/preprocess.py
```diff
@@ -0,0 +1,117 @@
...
+from spacy.tokens import Doc, DocBin
+
+DOCID_REGEX = "#begin document \((.*)\); part (\d*)"
+
+
+def read_file(fname, outname):
+    with open(fname, encoding="utf-8") as infile:
+        text = infile.read()
+
+    # TODO use larger model
```
svlandeg, line 19: Is this TODO still relevant?

experimental/coref/scripts/prep_span_data.py
```diff
@@ -0,0 +1,170 @@
...
+# the clustering component can be used. In tests with OntoNotes, accuracy and
+# processing time were similar for both options. However, silver heads avoid
+# tokenization mismatches, so they are chosen as the default.
+use_gold_heads = False
+if args.heads == "silver":
+    use_gold_heads = False
+elif args.heads == "gold":
+    use_gold_heads = True
+else:
+    raise ValueError(f"Expected 'gold' or 'silver' for heads but got: {args.heads}")
```
svlandeg, line 104: I'd avoid the indirect `if`-`else` logic to raise the error, and instead I'd write
```suggestion
if args.heads not in ["silver", "gold"]:
    raise ValueError(f"Expected 'gold' or 'silver' for heads but got: {args.heads}")
use_gold_heads = False
if args.heads == "gold":
    use_gold_heads = True
```

svlandeg (commented):
Had a few more smaller comments, then this should be good to merge?

experimental/coref/scripts/prep_span_data.py
```diff
@@ -0,0 +1,170 @@
...
+    for name, head_group in ex.predicted.spans.items():
+        cluster_id = name.split("_")[-1]
+        if not name.startswith(args.head_prefix):
+            continue
+        new_head_spangroup = []
+        new_span_spangroup = []
+        spans_name = f"{args.span_prefix}_{cluster_id}"
+        for head in head_group:
+            total_heads += 1
+            # Only one sample per head for the SpanPredictor
```
svlandeg, line 146:
```suggestion
            # Only one sample per head for the SpanResolver
```

experimental/coref/scripts/prep_span_data.py
```diff
@@ -0,0 +1,170 @@
+"""
+Kinda slow process mainly bottleneck by the
+prediction speed of the pipeline with the
+CorefScorer.
```
svlandeg, line 4:
```suggestion
CorefClusterer.
```

experimental/coref/test_project_coref.py
polm: Ah, you're right. I thought each test file was isolated, like a process, but that's not the case. I'll save and restore this value.

experimental/coref/scripts/preprocess.py
polm: No, it dates from when the small pipeline was used in development.

polm: I removed it.

experimental/coref/test_project_coref.py
polm: This should be fixed now.

experimental/coref/project.yml
```diff
@@ -0,0 +1,219 @@
...
+      branch: "main"
+      path: ""
+    description: "CoNLL-2012 scripts and dehydrated data, used for preprocessing OntoNotes."
+  - dest: assets/litbank
+    extra: True
+    git:
+      repo: "https://github.com/dbamman/litbank"
+      branch: "master"
+      path: "coref/conll"
+    description: "LitBank dataset. Only used for building data for tests."
```
svlandeg, line 93:
```suggestion
    description: "LitBank dataset. Only used for building data for quick unit tests."
```

experimental/coref/README.md
```diff
@@ -0,0 +1,108 @@
...
+### ⏯ Commands
+
+The following commands are defined by the project. They
+can be executed using [`spacy project run [name]`](https://spacy.io/api/cli#project-run).
+Commands are only re-run if their inputs have changed.
+
+| Command | Description |
+| --- | --- |
+| `prep-ontonotes-data` | Rehydrate the data using OntoNotes |
+| `prep-test-data` | Prepare minimal dataset for CI testing. Note this will overwrite train/dev/test data! |
```
svlandeg, line 75: Users might be interpreting "test" as "test data" rather than "unit test" here, so I wonder whether there's anyway we can make it more clear to users that they shouldn't really call this. I'd also consider using the explicit term "unit test" somewhere, in addition to "CI".

So maybe something like `prep-artificial-unit-test-data` (clunky on purpose) ?

experimental/coref/README.md
```diff
@@ -0,0 +1,108 @@
...
+The following workflows are defined by the project. They
+can be executed using [`spacy project run [name]`](https://spacy.io/api/cli#project-run)
+and will run the specified commands in order. Commands are only re-run if their
+inputs have changed.
+
+| Workflow | Steps |
+| --- | --- |
+| `prep` | `preprocess` |
+| `train` | `train-cluster` &rarr; `prep-span-data` &rarr; `train-span-resolver` &rarr; `assemble` |
+| `ci-test` | `prep-test-data` &rarr; `train-cluster` &rarr; `prep-span-data` &rarr; `train-span-resolver` &rarr; `assemble` &rarr; `eval` |
```
svlandeg, line 94: I would consider removing the `ci-test` workflow entirely and instead call the separate steps from within the unit test. Just so that this doesn't show up for users and doesn't potentially confuse them as to which workflow is most appropriate to them. I don't think we'd typically run this `ci-test` workflow from the commandline ourselves, right?

svlandeg (approved):
This has come together really nicely! Thanks for all your work on this Paul. I left a few final comments on the CI workflow setup (basically thinking about how to avoid confusion to users as much as possible). Other than that, I think this should be good to merge 🎉

experimental/coref/README.md
polm: That's a fair point. Since it's designed to be as similar to the real thing as possible it felt right to use the project file, but keeping it out of the way for users is more important.

experimental/coref/README.md
polm: Sounds good.

polm:
Since the test went green, I've gone ahead and updated the requirements to use v0.6.0 of `spacy-experimental` instead of the feature branch. The build will break until the release is actually made, but this will avoid another PR later.

polm:
Closing and re-opening to test.

PR #104, opened 2022-05-02
LEA only evaluation

kadarakos:
For simplicity I thought it makes sense to avoid having to raise the limit in the `conll` scripts, plus we can get rid of them. For this project the `lea` could be enough and we could decide later to add more evaluation metrics.


polm:
Thanks, looks good to me!

PR #105, opened 2022-05-04
Update tagger_parser_ud: add trainable lemmatizer, update for v3.3

adrianeboyd:


adrianeboyd:
The name is less and less accurate, but since it's used a fair bit in the docs I think it's better to keep the name and just extend the functionality to all the supported UD annotation (now everything except tokenization and MWT).

pipelines/tagger_parser_ud/configs/default.cfg
```diff
@@ -146,15 +174,16 @@ eps = 0.00000001
...
+dep_las = 0.08
 dep_las_per_type = null
 sents_p = null
 sents_r = null
-tag_acc = 0.33
-pos_acc = 0.17
-morph_acc = 0.17
-dep_uas = 0.17
-dep_las = 0.17
 sents_f = 0.0
```
svlandeg, line 187: These don't sum up to 1 anymore, which feels nicer? Maybe we can just have each weighted as 0.16?

pipelines/tagger_parser_ud/configs/default.cfg
```diff
@@ -146,15 +174,16 @@ eps = 0.00000001
 learn_rate = 0.001
 
 [training.score_weights]
+tag_acc = 0.16
+pos_acc = 0.08
+morph_acc = 0.08
 morph_per_feat = null
+lemma_acc = 0.08
+dep_uas = 0.08
+dep_las = 0.08
```
adrianeboyd, line 183:
```suggestion
tag_acc = 0.16
pos_acc = 0.16
morph_acc = 0.16
morph_per_feat = null
lemma_acc = 0.16
dep_uas = 0.16
dep_las = 0.16
```

PR #106, opened 2022-05-04
adjust paths in entity linker tutorial, fix file name

aplz:
This PR adjusts the paths specified in the sample command for being used directly after executing the workflow below.

Workflow:

```
python -m spacy project clone tutorials/nel_emerson
python -m spacy project assets /home/prodigy/nel_emerson
python -m spacy project run all
```


svlandeg (approved):
Thanks! I agree that it's more user-friendly to already specify the folders in which to find the various files that are needed.

PR #108, opened 2022-06-03
NEL benchmark

rmitsch:
## Goals

Add a benchmark for spaCy's NEL component.

## Description

Adds a new project to be used for benchmarking the F/P/R scores for spaCy's NEL component (in isolation from the NER step). The idea is to have a somewhat generic structure supporting the addition of new datasets. For now, only the [Reddit Entity Linking dataset](https://zenodo.org/record/3970806/) is supported.

The code also assumes that we are always and only using Wikidata as external knowledge base. If this is not the case, adjustments are necessary. Right now a few sentences of the first paragraph of the entity's page on Wikipedia are used to generate the entity description vector.

### Types of change

New feature/project.

## Todos

- [x] Test(s)
- [x] Auto-generate, manually extend project documentation
- [x] Typing improvements - create Pydantic schemas for complex types

## Checklist


benchmarks/nel/scripts/datasets/utils.py
```diff
@@ -0,0 +1,188 @@
...
+    entities (ENTITIES_TYPE): Entities to resolve.
+    entity_titles (Set[str]): List of entity titles to resolve.
+    batch_size (int): Number of entity titles to resolve in the same API request. Between 1 and 50.
+    RETURNS (Dict[str, str]): Mapping of titles to entity IDs.
+    """
+
+    assert 1 <= batch_size <= MAX_WIKI_API_BATCH_SIZE, \
+        f"Batch size has to be between 1 and {MAX_WIKI_API_BATCH_SIZE}."
+    entity_titles = list(entity_titles if entity_titles else entities.keys())
+    pbar = tqdm.tqdm(total=len(entity_titles))
```
svlandeg, line 47: The `tqdm` progress bar could use a title here so you know what it is exactly you're waiting for - something like "Processing relevant Wikidata entities"?

svlandeg: Or alternatively just a log statement to go before it. Right now the output is:
```

================================= create_kb =================================
Running command: '...\python.exe' ./scripts/create_kb.py reddit en_core_web_md
100%|██████████| 5167/5167 [01:42<00:00, 50.60it/s]
Trying to salvage 12 failed lookups
100%|██████████| 12/12 [00:02<00:00,  4.65it/s]
Constructing knowledge base with 5022 entries
```
So there's currently no explanation about the first progress bar.

svlandeg: And one final nitpick on that: it also finishes with the "Constructing ..." log statement. Maybe we want to add one more log statement noting it finished succesfully? Otherwise it might look like it crashed/stopped mid-through.

benchmarks/nel/scripts/datasets/dataset.py
```diff
@@ -0,0 +1,301 @@
...
+
+        root_path = Path(os.path.abspath(__file__)).parent.parent.parent
+        assets_path = root_path / "assets" / dataset_id
+
+        return {
+            "root": root_path,
+            "assets": assets_path,
+            "nlp_base": root_path / "temp" / f"{dataset_id}.nlp",
+            "nlp_best": root_path / "training" / dataset_id / "model-best",
+            "kb": root_path / "temp" / f"{dataset_id}.kb",
```
svlandeg, line 63: This might just be me, but I find dots in folder names pretty confusing - maybe let's use a dash or underscore instead?

benchmarks/nel/scripts/datasets/dataset.py
```diff
@@ -0,0 +1,301 @@
...
+        """ Parses external corpus. Loads data on entities and mentions.
+        Populates self._entities, self._failed_entity_lookups, self._annotations.
+        RETURNS (Tuple[ENTITIES_TYPE, Set[str], ANNOTATIONS_TYPE]): entities, titles of failed entity lookups,
+            annotations.
+        """
+        raise NotImplementedError
+
+    def _serialize_corpora(self) -> None:
+        """ Serializes corpora.
+        docs (List[Doc]): List of documents with entity annotations.
```
svlandeg, line 142:
```suggestion
```

benchmarks/nel/scripts/compile_corpora.py
```diff
@@ -0,0 +1,16 @@
...
+
+from datasets.dataset import Dataset
+
+
+def main(dataset_id: str):
+    """ Create corpora in spaCy format.
+    dataset_id (str): Dataset ID.
+    """
+
+    Dataset.generate_dataset_from_id(dataset_id).compile_corpora()
```
svlandeg, line 12: Is this creating the `Dataset` again from scratch?

svlandeg (commented):
This is looking good! I played around with it for a bit and everything was working fine up until the `run compile_corpora` command, which somehow seems flaky. It occassionally exists without having saved the `.spacy` files - I wonder whether  `spacy projects` is maybe swallowing an error message. Have you ran into this, or does this command always work for you? 


As a generic comment, check out the scripts we have to produce automated documentation for both the project as well as the parent directory (`benchmarks`): https://github.com/explosion/projects#%EF%B8%8Frepository-maintanance

benchmarks/nel/scripts/datasets/reddit.py
```diff
@@ -0,0 +1,149 @@
...
+    """ RedditEL dataset. """
+
+    @property
+    def dataset_id(self) -> str:
+        return "reddit"
+
+    def _parse_external_corpus(self) -> Tuple[ENTITIES_TYPE, Set[str], ANNOTATIONS_TYPE]:
+        file_names = [
+            f"{quality}_{source}_annotations.tsv"
+            for quality in ("gold", "silver", "bronze")
```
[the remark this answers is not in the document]
rmitsch: I missed adjusting that - will be done w.r.t. the dataset config in the next commit.

benchmarks/nel/scripts/datasets/dataset.py
rmitsch: Sure!

benchmarks/nel/scripts/compile_corpora.py
rmitsch: Yes. The creation of a `Dataset` instance is very lightweight however - the only thing happening there by default is setting the paths w.r.t. the dataset ID and loading the dataset config file.

benchmarks/nel/scripts/candidate_generation/embeddings.py
```diff
@@ -0,0 +1,32 @@
...
+from spacy.kb import KnowledgeBase, Candidate
+from spacy.tokens import Span
+
+from scripts.datasets.dataset import Dataset
+
+
+_pipelines: Dict[Path, Language] = {}
+
+
+def create_candidates(dataset_id: str, kb: KnowledgeBase, span: Span) -> Iterator[Candidate]:
```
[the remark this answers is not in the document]
rmitsch: Sure, I'll change it to `dataset_name` (I wouldn't call it just `dataset`, because for me that it implies it's actually a `Dataset` object).

benchmarks/nel/scripts/datasets/dataset.py
```diff
@@ -0,0 +1,301 @@
...
+        self._kb.initialize_entities(len(self._entities))
+        self._kb.initialize_vectors(len(self._entities))
+        entity_list: List[str] = []
+        freq_list: List[int] = []
+        vector_list: List[numpy.ndarray] = []
+        for qid, info in self._entities.items():
+            entity_list.append(qid)
+            freq_list.append(info["frequency"])
+            vector_list.append(self._nlp_base(info["description"]).vector)
+        self._kb.set_entities(entity_list=entity_list, vector_list=vector_list, freq_list=freq_list)
```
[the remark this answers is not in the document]
svlandeg: Yes that's exactly what they do :-) But we shouldn't expose that functionality to the user, you're right. Can you make a PR to spaCy to make them private?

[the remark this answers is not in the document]
svlandeg: Basically we really want to guide users to `set_entities` which will cover everything for them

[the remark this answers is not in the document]
rmitsch: Makes sense. PR: https://github.com/explosion/spaCy/pull/10978

rmitsch:
After the bumping `stanza` to v1.3+ the `stanza` model downloads succeed. The `textcat_architectures` project downloads resources directly from http://ai.stanford.edu though. I'll restart the tests once the site is up again.

benchmarks/nel/scripts/datasets/dataset.py
```diff
@@ -0,0 +1,300 @@
...
+        """ Returns dataset name.
+        """
+        raise NotImplementedError
+
+    def create_knowledge_base(self, model_name: str, **kwargs) -> None:
+        """ Creates and serializes knowledge base.
+        vectors_model (str): Name of model with word vectors to use.
+        """
+
+        self._nlp_base = spacy.load(model_name, exclude=["tagger", "lemmatizer"])
```
polm, line 81: What's the rationale for the exclude argument here? If it's for speed it would be best to exclude other components too, right?

benchmarks/nel/scripts/datasets/reddit.py
```diff
@@ -0,0 +1,157 @@
...
+                "`` How to Lose a Guy in 10 Days": "How to Lose a Guy in 10 Days",
+                "\"How to Lose a Guy in 10 Days": "\"How to Lose a Guy in 10 Days\"",
+                "Money\t325\t330\tmoney\n": "Money\t323\t328\tmoney\n"
+            }
+        }
+        encoding = "utf-8"
+
+        for file_name in to_replace:
+            lines: List[str] = []
+            with open(self._paths["assets"] / file_name, "r", encoding=encoding) as file:
```
polm, line 147: I think that utf-8 is the default and doesn't have to specified here in Python3?

benchmarks/nel/scripts/datasets/download_mewsli-9.sh
```diff
@@ -0,0 +1,32 @@
...
+
+# Downloads and sets up Mewsli-9 dataset.
+cd assets
+SUBDIR=dense_representations_for_entity_retrieval
+svn export https://github.com/google-research/google-research/trunk/$SUBDIR
+# Fix outdated requirement.
+sed -i -e 's/absl_py>=0.8.1/absl-py/g' $SUBDIR/requirements.txt
+# Only download and process English data.
+LANGS="ar de en es fa ja sr ta tr"
+LANG_LIST=($(echo $LANGS | tr ' ' "\n"))
```
polm, line 13: Minor point but do you need this variable? Since it's only used for iteration in bash I think it would behave the same as `$LANGS`?

polm (commented):
Sorry it took me so long to get to this! I have some questions and minor points but I think it looks pretty good. 

I had some issues running `spacy project run all`, I'm not sure if I'm doing something wrong. In the training step I get a `W093`, which suggests I have no NEL annotations. After that, in evaluation, I get `ModuleNotFoundError: No module named 'scripts'`. From the project directory I can `import scripts` just fine so I'm not sure if something is changing the working directory in a script or what. 

I think something may be funky with my virtualenv (left over from the previous PR or something?), so I'll clean it up and try again.

benchmarks/nel/README.md
```diff
@@ -0,0 +1,55 @@
...
+be fetched by running [`spacy project assets`](https://spacy.io/api/cli#project-assets)
+in the project directory.
+
+| File | Source | Description |
+| --- | --- | --- |
+| `assets/reddit.zip` | URL | Entity linking dataset scraped from Reddit. See [paper](https://arxiv.org/abs/2101.01228). |
+
+<!-- SPACY PROJECT: AUTO-GENERATED DOCS END (do not remove) -->
+
+Note that `svn` is required for downloading the Mewsli-9 dataset (should be replaceable by `git` with sparse
```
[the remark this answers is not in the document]
rmitsch: More of a todo for us. You're right that this is the wrong place for it. So `git` has a partial/sparse checkout feature that should work to the same effect, but so far I couldn't get it to work properly: commands that should work either error out or the entire `google-research` repo (at least without history) is downloaded. `svn export` is way faster in comparison.

I'll add this as a todo somewhere else and look into replacing `svn` with `git` at a later point.

benchmarks/nel/configs/nel.cfg
```diff
@@ -0,0 +1,151 @@
...
+before_creation = null
+after_creation = null
+after_pipeline_creation = null
+tokenizer = {"@tokenizers":"spacy.Tokenizer.v1"}
+
+
+[components]
+
+[components.senter]
+source = "en_core_web_md"
```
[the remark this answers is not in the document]
rmitsch: Not that I'm aware of. Will fix.

benchmarks/nel/scripts/datasets/reddit.py
rmitsch: I stumbled over some encoding issues with some of the text in the Reddit files which were fixed by setting UTF-8 explicitly.

benchmarks/nel/scripts/datasets/download_mewsli-9.sh
rmitsch: I decided to continue with the Mewsli-9 dataset in https://github.com/rmitsch/spacy-projects/pull/1. The newer version there drops this variable.
From what I remember, iterating over a string such as `$LANGS` behaves differently though than iterating over a list like `$LANG_LIST`.

benchmarks/nel/scripts/datasets/dataset.py
rmitsch: AFAIK (this is an Emerson snippet) this is only for speed reasons. Will drop other components too.

benchmarks/nel/README.md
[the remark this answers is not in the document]
polm: OK, got it. I think it's fine to not put too much effort into this if `git svn` doesn't seem to be working, we can always come back to it later.

benchmarks/nel/scripts/datasets/reddit.py
polm: Huh, I looked it up and UTF-8 is *not* the default - it uses the "preferred locale" set by the OS ([ref](https://docs.python.org/3.9/library/functions.html?highlight=open#:~:text=the%20default%20encoding%20is%20platform%20dependent). So that's fine.

rmitsch:
> Sorry it took me so long to get to this! I have some questions and minor points but I think it looks pretty good.
> 
> I had some issues running `spacy project run all`, I'm not sure if I'm doing something wrong. In the training step I get a `W093`, which suggests I have no NEL annotations. After that, in evaluation, I get `ModuleNotFoundError: No module named 'scripts'`. From the project directory I can `import scripts` just fine so I'm not sure if something is changing the working directory in a script or what.
> 
> I think something may be funky with my virtualenv (left over from the previous PR or something?), so I'll clean it up and try again.

No worries! And no, that's not your virtualenv. The current setup has some weird import issues - import statements either work in `train` or `evaluate`, but not in both. Added a workaround - I'd be happy to figure out the underlying issue though.

The `W093` is IIRC due to some documents not having any entities in them. I should filter those out, but in general the training process should execute well enough (most documents have entities).

benchmarks/nel/scripts/datasets/dataset.py
```diff
@@ -109,6 +106,7 @@ def create_knowledge_base(self, model_name: str, **kwargs) -> None:
         if not os.path.exists(self._paths["nlp_base"]):
             os.mkdir(self._paths["nlp_base"])
         self._nlp_base.to_disk(self._paths["nlp_base"])
+        print("Successfully constructed knowledge base.")
```
svlandeg, line 109: Let's replace these `print` statements with proper info logging calls

benchmarks/nel/scripts/datasets/dataset.py
```diff
@@ -0,0 +1,300 @@
...
+                    numpy.asarray(range(len(self._annotated_docs))),
+                    [
+                        int(self._options["frac_train"] * len(self._annotated_docs)),
+                        int((self._options["frac_train"] + self._options["frac_dev"]) * len(self._annotated_docs))
+                    ]
+                )
+            )
+        }
+
+        for key in indices:
```
svlandeg, line 158: We can iterate over 
```
for key, value in indices.items()
```
instead and then 
```
for idx in value
```

benchmarks/nel/scripts/datasets/utils.py
```diff
@@ -0,0 +1,188 @@
...
+) -> Tuple[List[Span], List[Dict[str, Union[Set[str], str, int]]]]:
+    """ Creates spans from annotations for one document.
+    doc (Doc): Document for whom to create spans.
+    entities_info (ENTITIES_TYPE): All available entities.
+    annotation (List[Dict[str, Union[Set[str], str, int]]]): Annotations for this post/comment.
+    entities_entities_failed_lookups (Set[str]): Set of entity names for whom Wiki API lookup failed).
+    source_id (str): Unique source ID to look up annotation.
+    RETURNS (Tuple[List[Span], List[Dict[str, Union[Set[str], str, int]]]]): List of doc spans for annotated entities;
+        list of overlapping entities.
+    """
```
svlandeg, line 142:
```suggestion
def _create_spans_from_doc_annotation(
    doc: Doc,
    entities_info: ENTITIES_TYPE,
    annotations: List[Dict[str, Union[Set[str], str, int]]],
    entities_failed_lookups: Set[str]
) -> Tuple[List[Span], List[Dict[str, Union[Set[str], str, int]]]]:
    """ Creates spans from annotations for one document.
    doc (Doc): Document for which to create spans.
    entities_info (ENTITIES_TYPE): All available entities.
    annotations (List[Dict[str, Union[Set[str], str, int]]]): Annotations for this post/comment.
    entities_failed_lookups (Set[str]): Set of entity names for which Wiki API lookup failed).
    RETURNS (Tuple[List[Span], List[Dict[str, Union[Set[str], str, int]]]]): List of doc spans for annotated entities;
        list of overlapping entities.
    """
```

benchmarks/nel/scripts/datasets/dataset.py
```diff
@@ -0,0 +1,300 @@
...
+        """
+
+        self._load_resource("entities")
+        self._load_resource("failed_entity_lookups")
+        self._load_resource("annotations")
+        self._load_resource("nlp_base")
+
+        Doc.set_extension("overlapping_annotations", default=None)
+        self._annotated_docs = self._create_annotated_docs()
+        self._serialize_corpora()
```
svlandeg, line 122: Should this have a succesful logging message after serializing? Maybe adding the location of where the corpora were stored?

svlandeg:
Actually! W093 is only thrown by the legacy `EntityLinker.v1` - we should upgrade to v2 for this project!

benchmarks/nel/scripts/datasets/evaluation.py
```diff
@@ -0,0 +1,207 @@
...
+            field_names=[
+                "model_name", "TPOS", "FPOS", "FNEG", "F-score", "Recall", "Precision",
+                *[f"F-score ({label})" for label in labels]
+            ]
+        )
+
+        for eval_result in evaluation_results:
+            eval_result._extend_report_table(table, tuple(labels))
+
+        print(table)
```
svlandeg, line 107: You need a pretty wide screen to be able to read this on the console.

I'd suggest printing the first bit separately (TPOS, FPOS, FNEG, F-SCORE, Recall, Precision) and then make a separate table which has each NE type on a separate row and for each you could print F/r/p

benchmarks/nel/scripts/datasets/evaluation.py
```diff
@@ -0,0 +1,207 @@
+""" Evaluation utilities.
+Adapated from https://github.com/explosion/projects/blob/master/nel-wikipedia/entity_linker_evaluation.py.
```
svlandeg, line 2:
```suggestion
Adapted from https://github.com/explosion/projects/blob/master/nel-wikipedia/entity_linker_evaluation.py.
```

svlandeg (approved):
This looks good to merge so I'll set this to approved. I know you probably want to add some tests yet first, but feel free to merge after that once the CI is green.

rmitsch:
@polm Good to merge from your side as well?

PR #111, opened 2022-06-22
Remove suggester part from spancat prodigy project

thomashacker:
This `PR` removes the suggester part from the project so that it's aligned with the upcoming video


svlandeg (approved):
It looks like the CI tests are failing due to connectivity/URL issues in some other example projects. We can still go ahead and get this merged if you like, but if you have time, could you have a look at the test failures, to see whether this was just a fluke or something else is going on?

PR #113, opened 2022-07-05
Run tests only in changed top-level directories

rmitsch:
## Goals

Run tests only for those projects changed in the latest commit.

## Description

Tests are run via a script extracting the latest changes with `git diff`. `pytest` is then called once for each modified top-level directory.

### Types of change

Enhancement.

## Checklist


adrianeboyd:
Can we do this by the second-level directories rather than the top-level directories?

run-tests.sh
```diff
@@ -0,0 +1,18 @@
...
+declare -a tested_dirs=()
+for dir in "${dirs[@]}"
+do
+  second_level_dir=$(echo "$dir" | awk -F/ '{print FS $2}')
+  second_level_dir="${second_level_dir:1}"
+
+  if [ ! -z "$second_level_dir" -a "$second_level_dir" != " " ]; then
+    if [[ ! " ${tested_dirs[*]} " =~ " ${second_level_dir} " ]]; then
+      tested_dirs+=($second_level_dir)
+      python -m pytest -s $second_level_dir
```
adrianeboyd, line 14: I think this needs to include the complete relative path to the second level directory?

rmitsch: Good catch. I messed up the git history,  clean PR/branch is [here](https://github.com/explosion/projects/pull/114).

PR #114, opened 2022-07-05
 Run tests only in changed second-level directories

rmitsch:
## Goals

Run tests only for those projects changed in the latest commit.

## Description

Tests are run via a script extracting the latest changes with `git diff`. `pytest` is then called once for each modified top-level directory.

### Types of change

Enhancement.

## Checklist


adrianeboyd:
This seems to work fine in my local tests and nothing major blows up even if it does turn out to need adjustments, so I think we should go ahead and merge it and just keep an eye on the CI output for the upcoming PRs to make sure it's doing what we expect.

PR #115, opened 2022-07-07
NEL: embedding-based & fuzzy lexical candidate selection

rmitsch:
## Goals

Implement a working draft picking entity candidates based on 
- similarities in the embedding space between entity description and text
- fuzzy lexical similarity between entity name and text

This description is WIP and will be updated.

## Description

Selects entity candidates based on similarities to the entity vectors in the embedding space, with some postprocessing.

### Types of change

New feature.

## Checklist


benchmarks/nel/configs/nel.cfg
```diff
@@ -39,10 +39,19 @@ entity_vector_length = 64
 incl_context = true
 incl_prior = true
 labels_discard = []
+;get_candidates = {"@misc":"spacy.CandidateGenerator.v1"}
+
+;[components.entity_linker.get_candidates]
+;@misc = "FuzzyStringGetCandidates.v1"
+;dataset_name = "${paths.dataset_name}"
+;max_n_candidates = 50
+;similarity_cutoff = .5
```
polm, line 48: Should this block be removed?

benchmarks/nel/project.yml
```diff
@@ -67,56 +77,80 @@ commands:
...
       - "assets/reddit/entities.pkl"
       - "assets/reddit/entities_failed_lookup.pkl"
       - "assets/reddit/annotations.pkl"
+#      - "assets/mewsli_9/entities.pkl"
+#      - "assets/mewsli_9/entities_failed_lookup.pkl"
+#      - "assets/mewsli_9/annotations.pkl"
     outputs_no_cache:
       - "temp/reddit/kb"
       - "temp/reddit/nlp"
+#      - "temp/mewsli_9/kb"
```
polm, line 94: Looks like there's a lot of commented code here that should be removed?

benchmarks/nel/scripts/parse_wiki_dumps.py
```diff
@@ -1,12 +1,26 @@
...
+    wiki_dump_api.parse(
+        entity_config={"limit": entity_limit},
+        article_text_config={"limit": article_limit},
+        alias_prior_prob_config={"limit": alias_limit}
+    )
 
 
 if __name__ == "__main__":
     typer.run(main)
+    # main(entity_limit=1000000, article_limit=1, alias_limit=1000000000)
```
polm, line 26: Comment can be removed? (Thanks for adding the limit param!)

benchmarks/nel/scripts/wiki/wikidata.py
```diff
@@ -3,21 +3,45 @@
...
+import io
 import json
+import os
 import sqlite3
 from pathlib import Path
-from typing import Union, Optional, Dict, Tuple, Any, List, Set
+from typing import Union, Optional, Dict, Tuple, Any, List, Set, Iterable, Iterator
+import time
+
+import indexed_bzip2
```
polm, line 14: I think this should be removed?

benchmarks/nel/project.yml
```diff
@@ -48,10 +55,13 @@ commands:
     script:
       - "unzip -q -d assets/reddit assets/reddit.zip"
       - "python ./scripts/clean_data.py reddit"
+      - "python ./scripts/clean_data.py mewsli-9"
```
polm, line 58: I think this should be `mewsli_9`? Got an import related error if I didn't change it.

benchmarks/nel/project.yml
```diff
@@ -48,10 +55,13 @@ commands:
     script:
       - "unzip -q -d assets/reddit assets/reddit.zip"
```
polm, line 56: Please add `-o` to always overwrite or something else so repeated runs don't ask for input.

polm (dismissed):
Still running this, but ran into a few minor issues that need fixing.

rmitsch:
> Still running this, but ran into a few minor issues that need fixing.

Thanks for the feedback! I'm currently working on the Mewsli-9 dataset - will clean up when that's done (and aim to merge afterwards).

benchmarks/nel/configs/nel.cfg
```diff
@@ -39,19 +39,19 @@ entity_vector_length = 64
...
 dataset_name = "${paths.dataset_name}"
-max_n_candidates = 100
-lexical_similarity_cutoff = .5
+max_n_candidates = 50
+similarity_cutoff = .5
+; ---------
+;@misc = "EmbeddingGetCandidates.v1"
+;dataset_name = "${paths.dataset_name}"
+;max_n_candidates = 100
+;lexical_similarity_cutoff = .5
```
polm, line 54: Should these be removed?

benchmarks/nel/scripts/compile_corpora.py
```diff
@@ -12,4 +12,5 @@ def main(dataset_name: str):
 
 
 if __name__ == "__main__":
-    typer.run(main)
+    # typer.run(main)
+    main("mewsli_9")
```
polm, line 16: Is this intentional, or is it a result from testing?

polm (commented):
Looks OK, a few small points.

benchmarks/nel/configs/nel.cfg
rmitsch: Yes, these are leftovers from testing.

benchmarks/nel/scripts/compile_corpora.py
rmitsch: The latter.

rmitsch:
> Couple of small things:

Sorry I didn't get back to you earlier! Didn't this last comment. Will clean up `project.yml` and the readme.

rmitsch:
On a different note: running the entire project is obviously unfeasible since the addition of the Wiki dump download and parsing. Other steps could be limited in their runtime rather easily, but we'd still need the downloaded complete Wiki dump, unfortunately. Unsure as to how to proceed here. Any suggestions?

polm:
Can we directly provide the output of processing the dataset somehow? If hosting isn't difficult that would be easiest to set up.

A more principled thing would be to pick a subset, like "people born after 1900" or something. Picking a reasonable subset could be hard though.

rmitsch:
> Can we directly provide the output of processing the dataset somehow? If hosting isn't difficult that would be easiest to set up.
> 
> A more principled thing would be to pick a subset, like "people born after 1900" or something. Picking a reasonable subset could be hard though.

Yes, we could host the .sqlite file with the parsed data. In total it's around 14 GB. I'll look into figuring out a reasonable subset.

rmitsch:
On second thought, this would bypass testing the parsing script. It'd be better if we started with small Wiki dumps and went from there. One option is to 
- locally run through both dumps, 
- pick a limited number of entries - either randomly or in a more principled fashion,
- copy them into a new archive,
- upload the archive so it's publicly available,
- in the test suite: download the smaller test dumps and run all steps (I guess I'd have to adjust `project.yml` for the test case).

This seems like a bother, but is perhaps the only way to test the entire project. Except @svlandeg has a better idea?

svlandeg:
I'm quite fond of having an alternative, much smaller dataset, even if it's perhaps unrealistically small, for testing purposes. Not just for the CI, but also to allow people to have a quick look & run through the project without first having to sit around for hours to wait for the download to finish.

We could think of a very narrow domain like Paul mentioned, and then your outline Raphael does sound sensible to me!

rmitsch:
Suggestion: 
- Pick every Wiki entity mentioning at least one of a set of seed terms related to astronomy (moon, star, galaxy, ...) in their description.
- Parse the Wiki dump again (:smiling_face_with_tear:) and move these entities + their articles into two separate entities/articles archives.
- Upload the archives (where to?) and make them publicly available.
- Comment the original Wiki dump download links and replace them (with an appropriate notice) with links to the smaller test sets.

We could do something similar for the Mewsli dataset, but IIRC the legal situation around redistributing the data was unclear. We could download it in full and then filter it afterwards though with similar terms, otherwise we'd have a large non-astronomy corpus and only astronomical entities to link.

polm:
I think filtering based on seed terms like that makes sense. 

> Upload the archives (where to?) and make them publicly available.

I think the default for this would usually be Github Releases, unless the data is too large or something.

rmitsch:
I filtered by a list of European capitals now (testing the mentioned teams/city names idea). The filtered Wikidata and Wikipedia dumps are 74 and 58 MB respectively. Are we ok with that or do we want a smaller set?

rmitsch:
Git LFS is another option, I guess? Feels kinda weird to me if the only "release" in the repo were the Wiki dump.

polm:
Sorry for the late reply. 

> I filtered by a list of European capitals now (testing the internally discussed sports teams/city names idea). The filtered Wikidata and Wikipedia dumps are 74 and 58 MB respectively. Are we ok with that or do we want a smaller set?

Those sizes sounds great, easy to work with. 

> Git LFS is another option, I guess? Feels kinda weird to me if the only "release" in the repo were the Wiki dump.

We've done this for things like `sense2vec` before. Git LFS would also work but I think using it is pretty awkward, especially for data a user is only likely to use once, especially if we aren't very concerned about diffs.

For files <100MB you can just put them in git, though Github will warn that you should use LFS. We wouldn't want to do that in the main branch here though. One other thing I've heard of people doing is having a "data" branch, so normal checkouts are fast but data is still in git, though I think releases are still preferable since this won't be updated often.

benchmarks/nel/project.yml
```diff
@@ -89,9 +86,9 @@ commands:
       - "temp/${vars.dataset}/nlp"
 
   - name: compile_corpora
-    help: "Compile corpora, separated in in train/dev/test sets"
+    help: "Compile corpora, separated in in train/dev/test sets."
     script:
-      - "python ./scripts/compile_corpora.py ${vars.dataset}"
+      - "python ./scripts/compile_corpora.py ${vars.dataset} True"
```
polm, line 91: Since you're using Typer can't you use a flag here?

```suggestion
      - "python ./scripts/compile_corpora.py ${vars.dataset} --use-filter-terms"
```

benchmarks/nel/scripts/datasets/mewsli_9.py
```diff
@@ -68,12 +68,26 @@ def _create_annotated_docs(self) -> List[Doc]:
...
                     ) as text_file:
+                        lines = text_file.readlines()
+                        doc_text = " ".join(
+                            " ".join(
+                                [
+                                    line.replace("\n", " ").replace("\xa0", ".")
+                                    for line in lines
+                                ]
+                            ).split()
+                        )
```
polm, line 79: What is this code supposed to do? Also is it correct to replace `\xa0` (non-breaking space) with a period and not a space?

benchmarks/nel/scripts/wiki/wiki_dump_api.py
```diff
@@ -24,17 +28,41 @@ def establish_db_connection() -> sqlite3.Connection:
...
+    """Extracts small demo dump by parsing the Wiki dumps and keeping only those entities (and their articles)
+    containing any of the specified filter_terms. The retained entities and articles are written into
+    filter_terms (Set[str]): Terms having to appear in entity descriptions in order to be wrr
+    """
+
+    # entity_ids, entity_labels = wikidata.extract_demo_dump(
+    #     _paths["wikidata_dump"], _paths["filtered_wikidata_dump"], filter_terms
+    # )
+    # with open(_paths["filtered_entity_entity_info"], "wb") as file:
+    #     pickle.dump((entity_ids, entity_labels), file)
```
polm, line 41: If you're leaving this in for debugging or whatever that's fine, just going to flag it here in the review in case. I think there's a few more of these.

benchmarks/nel/scripts/wiki/wiki_dump_api.py
```diff
@@ -24,17 +28,41 @@ def establish_db_connection() -> sqlite3.Connection:
...
+
+    with open(_paths["filtered_entity_entity_info"], "rb") as file:
+        _, entity_labels = pickle.load(file)
+    wikipedia.extract_demo_dump(_paths["wikipedia_dump"], _paths["filtered_wikipedia_dump"], entity_labels)
+
+    # try:
+    #     os.remove(_paths["filtered_entity_entity_info"])
+    # except OSError:
+    #     pass
```
polm, line 50: More comments.

benchmarks/nel/scripts/wiki/wikidata.py
```diff
@@ -102,6 +102,12 @@ def read_entities(
                 clean_line = line.strip()
                 if clean_line.endswith(b","):
                     clean_line = clean_line[:-1]
+                to_search = clean_line.decode("utf-8").lower()
+                if "boston" in to_search and "celtics" in to_search:
+                    obj = json.loads(clean_line)
+                    obj_lbl = obj["labels"].get("en", {}).get("value", "").lower()
+                    if "celtics" in obj_lbl:
+                        x = 3
```
polm, line 110: What is the purpose of this section? 

Also it might be better to call `json.loads` pre-emptively if this block is entered a lot, it looks like it can happen twice for the same line.

polm (commented):
Thank you for building out the filtered version of this! I had a few questions / comments.

benchmarks/nel/scripts/wiki/wiki_dump_api.py
rmitsch: Argh, commented that for testing. Will uncomment. Thanks!

benchmarks/nel/scripts/wiki/wiki_dump_api.py
rmitsch: Same.

benchmarks/nel/scripts/wiki/wikidata.py
rmitsch: Even more testing code I forgot to remove. I was tracking why a [particular entity](https://www.wikidata.org/wiki/Q131371) wasn't included in the filtered dataset.

benchmarks/nel/project.yml
rmitsch: Correct, replaced that meanwhile.

benchmarks/nel/scripts/datasets/mewsli_9.py
rmitsch: `\xa0` seems to be only used after a title in the Mewsli-9/Wikinews corpus, and there usually isn't any end-of-sentence punctuation after the title. I thought that converting `\xa0` into a a period instead might help the pipeline with parsing the text correctly.

This snippet, after replacing special characters with spaces/periods, condenses multiple to single whitespaces by splitting and then joining the document text.

benchmarks/nel/requirements.txt
```diff
@@ -1,3 +1,7 @@
 pyyaml
+requests
```
polm, line 2: Is requests actually used anywhere?

benchmarks/nel/requirements.txt
```diff
@@ -1,3 +1,7 @@
...
 tqdm
-prettytable
\ No newline at end of file
+prettytable
+scikit-learn
+fuzzyset2
+spacyfishing
```
polm, line 7: `datasets` seems to be required but is not included here?

polm (commented):
Still looking at everything but had some more comments about requirements.

benchmarks/nel/requirements.txt
rmitsch: No, that was used for querying the Wiki API, which has been replaced by parsing the dump locally. Removed.

benchmarks/nel/requirements.txt
rmitsch: Hm, why do you think it's required? `datasets` is a local package (`nel.scripts.datasets`).

polm: I got this error running `preprocess`:

```
================================= preprocess =================================
Running command: unzip -q -o -d assets/reddit assets/reddit.zip
Running command: /mnt/pool/code/spacy-projects/benchmarks/nel/env/bin/python ./scripts/clean_data.py nel.cfg
Traceback (most recent call last):
  File "/mnt/pool/code/spacy-projects/benchmarks/nel/./scripts/clean_data.py", line 20, in <module>
    typer.run(main)
  File "/mnt/pool/code/spacy-projects/benchmarks/nel/env/lib/python3.10/site-packages/typer/main.py", line 864, in run
    app()
  File "/mnt/pool/code/spacy-projects/benchmarks/nel/env/lib/python3.10/site-packages/typer/main.py", line 214, in __call__
    return get_command(self)(*args, **kwargs)
  File "/mnt/pool/code/spacy-projects/benchmarks/nel/env/lib/python3.10/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/mnt/pool/code/spacy-projects/benchmarks/nel/env/lib/python3.10/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/mnt/pool/code/spacy-projects/benchmarks/nel/env/lib/python3.10/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/mnt/pool/code/spacy-projects/benchmarks/nel/env/lib/python3.10/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/mnt/pool/code/spacy-projects/benchmarks/nel/env/lib/python3.10/site-packages/typer/main.py", line 500, in wrapper
    return callback(**use_params)  # type: ignore
  File "/mnt/pool/code/spacy-projects/benchmarks/nel/./scripts/clean_data.py", line 16, in main
    Dataset.generate_dataset_from_id(dataset_name).clean_assets()
  File "/mnt/pool/code/spacy-projects/benchmarks/nel/scripts/datasets/dataset.py", line 416, in generate_dataset_from_id
    importlib.import_module(module_name), inspect.isclass
  File "/usr/lib/python3.10/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 992, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1004, in _find_and_load_unlocked
ModuleNotFoundError: No module named 'datasets.nel'
```

And installing `datasets` made it go away, though I guess that was just an accident because of the name collision? Further steps in the pipeline ran OK afterwards too.

rmitsch: Thanks for the log! That was actually a misconfiguration in `project.yml` that should be fixed now. Can you give it another shot without `datasets`?

rmitsch:
Closing and reopening to trigger CI.

polm (approved):
Ran `spacy project run all` with no errors or issues - looks good!

benchmarks/nel/project.yml
```diff
@@ -4,27 +4,38 @@ vars:
...
 assets:
-  - dest: 'assets/reddit.zip'
-    url: 'https://zenodo.org/record/3970806/files/reddit_el.zip?download=1'
-    checksum: 'eea345cc7574a5c9c376748d1871d557'
-    description: Entity linking dataset scraped from Reddit. See [paper](https://arxiv.org/abs/2101.01228).
   - dest: 'assets/wiki/wikidata_entity_dump.json.bz2'
     url: 'https://dumps.wikimedia.org/wikidatawiki/entities/latest-all.json.bz2'
-    description: Wikidata entity dump.
+    description: Wikidata entity dump. Download can take a long time!
+    extra: True
```
svlandeg, line 19: This is a great solution to "hide" the full dumps behind the "extra" flag!

benchmarks/nel/project.yml
```diff
@@ -4,27 +4,38 @@ vars:
...
+    checksum: 'ba2d979105abf174208608b942242fcb'
+  - dest: 'assets/wiki/wikipedia_dump_filtered.xml.bz2'
+    url: 'https://github.com/explosion/projects/releases/download/nel-benchmark-filtered-wiki-data/wikipedia_dump_filtered.xml.bz2'
+    description: Filtered Wikipedia dump for demo purposes.
+    checksum: 'cb624eaa5887fe1ff47a9206c9bdcfd8'
 
 workflows:
   all:
-    - setup
+    - download_mewsli-9
```
svlandeg, line 35: I always find it awkward to type/remember commands that include both a `_` as well as a `-`. Can we maybe shorten this to `download_mewsli`?

rmitsch: Renamed it to `download_mewsli9`. If you prefer `download_mewsli`, feel free to reopen this conversation and I'll adjust it.

benchmarks/nel/README.md
```diff
@@ -47,9 +49,18 @@ in the project directory.
...
+| `assets/wiki/wikipedia_dump.xml.bz2` | URL | Wikipedia dump. Download can take a long time! |
+| `assets/wiki/wikidata_entity_dump_filtered.json.bz2` | URL | Filtered Wikidata entity dump for demo purposes. |
+| `assets/wiki/wikipedia_dump_filtered.xml.bz2` | URL | Filtered Wikipedia dump for demo purposes. |
 
 <!-- SPACY PROJECT: AUTO-GENERATED DOCS END (do not remove) -->
 
-Note that `svn` is required for downloading the Mewsli-9 dataset (should be replaceable by `git` with sparse 
-checkout).
\ No newline at end of file
+Notes:
```
svlandeg, line 59: There should be a very clear & visible warning that this currently doesn't work on Windows

rmitsch: Added one. If you think it's insufficient/would like to change something, feel free to reopen the conversation.

benchmarks/nel/project.yml
```diff
@@ -36,96 +47,94 @@ workflows:
...
       - "configs/datasets.yml"
     outputs_no_cache:
-      - "corpora/reddit/train.spacy"
-      - "corpora/reddit/dev.spacy"
-      - "corpora/reddit/test.spacy"
+      - "corpora/${vars.dataset}/train.spacy"
+      - "corpora/${vars.dataset}/dev.spacy"
+      - "corpora/${vars.dataset}/test.spacy"
 
   - name: train
```
svlandeg, line 109: I do get an error about data not being properly formatted, is this a worry? It does seem to start training afterwards so I'm not sure when exactly that warning is thrown or whether it's significant.

```
E    #       LOSS ENTIT...  SENTS_F  SENTS_P  SENTS_R  ENTS_F  ENTS_P  ENTS_R  NEL_MICRO_F  NEL_MICRO_R  NEL_MICRO_P  SCORE
---  ------  -------------  -------  -------  -------  ------  ------  ------  -----------  -----------  -----------  ------
...\spacy\pipeline\legacy\entity_linker.py:193: UserWarning: [W093] Could not find any data to train the Entity Linker on. Is your input data correctly formatted?
  warnings.warn(Warnings.W093.format(name="Entity Linker"))
  0       0           0.00    88.50    91.58    85.62    0.00    0.00    0.00        98.46        98.46        98.46    0.62
  0     200          23.64    88.50    91.58    85.62    0.00    0.00    0.00        98.46        98.46        98.46    0.62
  0     400          11.84    88.50    91.58    85.62    0.00    0.00    0.00        98.46        98.46        98.46    0.62
```

benchmarks/nel/project.yml
```diff
@@ -36,96 +47,94 @@ workflows:
...
+      - "bash scripts/train.sh ${vars.dataset} ${vars.config} ${vars.training_max_steps} ${vars.gpu_id} "
     deps:
-      - "temp/reddit/kb"
-      - "temp/reddit/nlp"
-      - "corpora/reddit/train.spacy"
-      - "corpora/reddit/dev.spacy"
+      - "temp/${vars.dataset}/kb"
+      - "temp/${vars.dataset}/nlp"
+      - "corpora/${vars.dataset}/train.spacy"
+      - "corpora/${vars.dataset}/dev.spacy"
```
svlandeg, line 117: The `outputs` specification is missing for the `train` command

benchmarks/nel/scripts/datasets/evaluation.py
```diff
@@ -144,8 +149,8 @@ def report(evaluation_results: Tuple["EvaluationResults"]) -> None:
             eval_result._extend_report_overview_table(overview_table)
             eval_result._extend_report_labels_table(label_table, tuple(labels))
 
-        logger.info(overview_table)
-        logger.info(label_table)
+        logger.info("\n" + str(overview_table))
+        logger.info("\n" + str(label_table))
```
svlandeg, line 153: This is from an older PR, but I would rename "TPOS", "FPOS" and "FNEG" to "TP", "FP" and "FN" as those terms are more standard.

Also, for the second table, is the main use-case that we want to compare performance on labels within a model, or do we want to compare the performance of one label across models? The table supports the former, but I was kind of expecting the latter, in which case I would sort the entries differently: first for each label show all models and their scores, only then continue to the next label. But it depends on what the main usage of this table is.

benchmarks/nel/scripts/datasets/dataset.py
```diff
@@ -220,61 +249,86 @@ def evaluate(
         candidate_generation: bool = True,
         baseline: bool = True,
         context: bool = True,
+        spacyfishing: bool = True,
         n_items: Optional[int] = None,
     ) -> None:
         """Evaluates trained pipeline on test set.
         baseline (bool): Whether to include baseline results in evaluation.
         context (bool): Whether to include the local context in the model.
+        spacyfishing (bool): Whether to include evaluation with spacyfishing.
```
svlandeg, line 258: The docstring is missing the `candidate_generation` part.

benchmarks/nel/configs/nel.cfg
```diff
@@ -41,8 +41,10 @@ incl_prior = true
 labels_discard = []
 
 [components.entity_linker.get_candidates]
-@misc = "EmbeddingGetCandidates.v1"
+@misc = "FuzzyStringGetCandidates.v1"
```
svlandeg, line 44: It's kind of unfortunate that the project doesn't show the difference between these two candidate generators in the `evaluate` step. I wonder whether that could be done by dynamically looping through the candidate generator options/configs, replacing them in the config at inference time and then show more detailed results instead of just "Candidate Gen." in the final tables?

That's assuming that we're training on gold entities, because otherwise the training is also affected by this choice.

benchmarks/nel/scripts/candidate_generation/lexical.py
```diff
@@ -0,0 +1,32 @@
+""" Candidate generation via distance in lexical space. """
+from typing import Iterable
+
+from spacy.kb import KnowledgeBase, Candidate
+from spacy.tokens import Span
+from .base import NearestNeighborCandidateSelector
+from cfuzzyset import cFuzzySet as FuzzySet
+
+
+class CandidateSelector(NearestNeighborCandidateSelector):
```
svlandeg, line 10: They're both called `CandidateSelector`?

benchmarks/nel/scripts/datasets/dataset.py
```diff
@@ -220,61 +249,86 @@ def evaluate(
         candidate_generation: bool = True,
         baseline: bool = True,
         context: bool = True,
+        spacyfishing: bool = True,
```
svlandeg, line 252: What does this do exactly?

benchmarks/nel/scripts/datasets/dataset.py
```diff
@@ -220,61 +249,86 @@ def evaluate(
...
         candidate_results = evaluation.EvaluationResults("Candidate gen.")
 
         for example in tqdm.tqdm(
             test_set, total=n_items, leave=False, desc="Processing test set"
         ):
+            example: Example
             if len(example) > 0:
-                correct_ents = {
+                entity_linker: EntityLinker_v1 = self._nlp_best.components[  # type: ignore
+                    self._nlp_best.component_names.index("entity_linker")
```
svlandeg, line 303: Lets move this index to a separate var so that hopefully the declaration of `entity_linker` can be one line instead.

svlandeg (commented):
Nice work! 

I've only tried running it on the filtered data, I suppose the high `Random` F-score (88%) is because the filtered data has much less ambiguity than a realistic dump would have? It makes sense to run a true benchmark on the unfiltered dataset though, so that's fine.

Have you run this on the full unfiltered datasets? What were the results? How do the different candidate generators compare?

rmitsch:
> Have you run this on the full unfiltered datasets? What were the results? How do the different candidate generators compare?

Not in the current version, but it's on my todo list.

benchmarks/nel/scripts/datasets/dataset.py
rmitsch: `spacyfishing` fetches its result individually via an API. This slows down the evaluation process by a lot. For the sake of convenience (esp. with bigger eval. sets) I added this flag in case I just wanted to compare our internal implementations.

benchmarks/nel/project.yml
```diff
@@ -36,96 +47,94 @@ workflows:
...
     script:
-      - "python -m spacy train configs/${vars.config} \
-          --paths.dataset_name reddit \
-          --output training/reddit \
-          --paths.train corpora/reddit/train.spacy \
-          --paths.dev corpora/reddit/dev.spacy \
-          --paths.kb temp/reddit/kb \
-          --paths.base_nlp temp/reddit/nlp \
-          -c scripts/custom_functions.py"
+      - "bash scripts/train.sh ${vars.dataset} ${vars.config} ${vars.training_max_steps} ${vars.gpu_id} "
```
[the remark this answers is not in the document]
rmitsch: And the Mewsli download fails? What's the output there, out of curiosity?

benchmarks/nel/project.yml
```diff
@@ -36,96 +47,94 @@ workflows:
...
     script:
-      - "unzip -q -d assets/reddit assets/reddit.zip"
-      - "python ./scripts/clean_data.py reddit"
+      - "python ./scripts/clean_data.py ${vars.dataset}"
     deps:
-      - "assets/reddit.zip"
+      - "assets/${vars.dataset}"
     outputs:
-      - "assets/reddit"
+      - "assets/${vars.dataset}"
```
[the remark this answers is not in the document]
rmitsch: We could introduce a new directory for the cleaned corpora. Didn't do that because I thought it'd be more elegant to modify the corpus in-place. Should we do a separate directory with a cleaned corpus copy?

benchmarks/nel/project.yml
rmitsch: This seems to be due to some documents not having entities in them (that happening once is enough to trigger the warning IIRC). So AFAIK this is not super relevant - when I checked most documents _had_ entities.

benchmarks/nel/project.yml
rmitsch: Added. Btw, is there a good reason to use `output_no_cache` for anything we produce here? Just noticed that I'm using `output_no_cache` in most steps, but don't remember why :upside_down_face:

benchmarks/nel/configs/nel.cfg
rmitsch: I was thinking about that too, but discarded the idea because I thought it would make things messier than necessary by moving away from the one config -> one run idea. Not ruling it out though. 

Alternative idea: store evaluation data in a machine-readable format and add a step comparing the performance from multiple collected runs somehow? 
So we'd have to run the whole workflow multiple times and then could do `spacy project run compare_evaluations` - or something similar - and get a table comparing the results from different runs (e.g. different candidate generators).

benchmarks/nel/scripts/candidate_generation/lexical.py
rmitsch: I kinda like the _different module name_ with _same class name_ pattern if the classes have the same purpose I'm good with changing the names though. `lexical.CandidateSelector` to `lexical.LexicalNNCandidateSelector`, `embeddings.CandidateSelector` -> `embeddings.EmbeddingNNCandidateSelector`?

benchmarks/nel/scripts/datasets/dataset.py
rmitsch: Dunno why it took me so long to learn about `.get_pipe()` :smile:

benchmarks/nel/project.yml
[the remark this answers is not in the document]
svlandeg: The Mewsli download failed because of `wget` not being available in Git Bash, but there will likely be other errors as well.

benchmarks/nel/project.yml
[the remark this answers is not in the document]
svlandeg: I'd personally like that better, but I guess it's a matter of taste, so I don't mind strongly either way.

benchmarks/nel/project.yml
svlandeg: Maybe we need to revisit when to raise this warning then. We currently raise it per batch on `update`, but maybe there's an earlier point in the pipeline when we could check whether there is *any* correct data, to avoid FP warnings like these.

benchmarks/nel/project.yml
svlandeg: cf https://spacy.io/usage/projects#deps-outputs :-)

benchmarks/nel/scripts/candidate_generation/lexical.py
svlandeg: I personally think it complicates discussions when talking about these things and you always have to contextualise the class names with the exact path. But yea the long names get very verbose :'(

benchmarks/nel/configs/nel.cfg
svlandeg: Yea, that'd be nice. And we could just store the different configs rather than automatically generating them, maybe, because then the configs can be more easily copy/pasted by users.

benchmarks/nel/scripts/candidate_generation/lexical.py
rmitsch: True. I'm good with changing the names, I think in this case the more verbose names are still ok. Should we?

benchmarks/nel/configs/nel.cfg
rmitsch: Makes sense for a benchmark, I think. I'd suggest doing that in a different PR though, if that's ok with you?

benchmarks/nel/scripts/datasets/evaluation.py
rmitsch: Haven't really thought about that - I agree that the latter use case makes more sense. Adapted it accordingly & renamed TPOS etc.

benchmarks/nel/scripts/candidate_generation/lexical.py
svlandeg: I think I'd drop the `NN` bit from the longer names, but yea otherwise they're OK to use for this project :-)

benchmarks/nel/configs/nel.cfg
svlandeg: Sure!

benchmarks/nel/project.yml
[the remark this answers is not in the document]
rmitsch: Added `raw` and `clean` directories in `/assets/${vars.dataset}`.

svlandeg (approved):
Nice work, Raphael!

PR #117, opened 2022-07-18
Extend speed benchmark requirements to avoid sklearn import exception

rmitsch:
## Goals

Fix the [`__check_build` import error for `scikit-learn` in the speed benchmark](https://buildkite.com/explosion-ai/spacy-projects/builds/24#01821079-7b9c-4a44-a5af-73cc758620b7/112-366). 

## Description

Add `scipy` as explicit dependency. 

### Types of change

Bug fix.

## Checklist

### DIFF
--- a/benchmarks/speed/requirements.txt
+++ b/benchmarks/speed/requirements.txt
@@ -1,5 +1,8 @@
 transformers>=3.4.0,<4.3.0
 spacy-transformers>=1.0.1,<1.1.0
+numpy>=1.19.0,<1.19.6
+scipy>=1.7.0,<1.7.4
+scikit-learn>=1.0.0,<1.0.3
 stanza>=1.3.0,<1.5.0
 flair>=0.6.0,<1.0.0
 ufal.udpipe
--- a/benchmarks/speed/requirements.txt
+++ b/benchmarks/speed/requirements.txt
@@ -1,8 +1,6 @@
+scipy>=1.7.0,<1.7.4
 transformers>=3.4.0,<4.3.0
 spacy-transformers>=1.0.1,<1.1.0
-numpy>=1.19.0,<1.19.6
-scipy>=1.7.0,<1.7.4
-scikit-learn>=1.0.0,<1.0.3
 stanza>=1.3.0,<1.5.0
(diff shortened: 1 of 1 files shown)


adrianeboyd:
This change isn't a huge deal, but where is this actually going wrong? It looks like scipy 1.7.3 is installed by the buildkite script, so why is this failing later?

(If we're moving from running the tests for individual directories we should switch from installing requirements globally in the CI to installing from `requirements.txt` for each project, I think. It shouldn't be necessary that there is one set of global requirements that works for all projects and it would help test that the requirements stay correct. Then maybe we could also move away from this old torch version for most newer projects.)

rmitsch:
- Global requirements: I agree. Added a ticket.
- `scipy` is currently being installed by the BuildKite script to avoid this error. The plan is to remove this line as soon as this is merged (exactly for that reason: it's not a global requirement, it's needed only by the speed benchmark).

adrianeboyd:
But why do you see this failure with this version of scipy already installed?

adrianeboyd:
Well, I ran into an unexpected torch+numpy bug (in certain version combinations) testing this that we need to have a look at in `thinc`.

Let me try updating this project separately and testing the new modified-project-only test script...

adrianeboyd:
So I'm not sure what's going on with the test run above, but I don't think we need to add `scipy` as a separate dependency here. It does look like `benchmarks/speed` could use some updates in general, which I tested a bit in combination with some minor changes to the CI setup in the testing-only PR #118.

rmitsch:
Sorry for not getting back to you earlier. So, this is weird - I ran [the job again](https://buildkite.com/explosion-ai/spacy-projects/steps), this time without the explicit `scipy` installation, and it passed. So it does seem like `scipy` is not needed to be installed explicitly, but I have no idea why it failed in the first place (some [digging](https://stackoverflow.com/questions/15274696/importerror-in-importing-from-sklearn-cannot-import-name-check-build) indicates that installing `scipy` might be a solution for this issue).

I'll close this PR and keep an eye on the weekly builds - if the problem occurs again, I'll have another look. Will check out https://github.com/explosion/projects/pull/118 once it's no longer WIP.

PR #119, opened 2022-07-20
Install per-project requirements in CI tests

adrianeboyd:


adrianeboyd:
No project tests are triggered here because only the CI code is modified, but the same changes passed with a modified project in #118.

PR #120, opened 2022-07-20
Update speed benchmark

adrianeboyd:
Update speed benchmark for newer `spacy-transformers` and `flair` APIs.

Requires #119 for the tests to pass.

### DIFF
--- a/benchmarks/speed/project.yml
+++ b/benchmarks/speed/project.yml
@@ -26,7 +26,6 @@ directories: ["scripts", "texts", "results"]
 # haven't changed, it won't be re-run.
 workflows:
   setup:
-    - install
     - download
   benchmark:
     - timing_cpu
@@ -43,13 +42,6 @@ assets:
 # via "spacy project run [command] [path]". The help message is optional and
 # shown when executing "spacy project run [optional command] [path] --help".
 commands:
-  - name: install
-    help: "Install dependencies"
-    script:
-      - "python -m pip install -r requirements.txt -f https://download.pytorch.org/whl/torch_stable.html"
-    deps:
-      - "requirements.txt"
-
   - name: download
     help: "Download models"
     script:
--- a/benchmarks/speed/project.yml
+++ b/benchmarks/speed/project.yml
@@ -26,6 +26,7 @@ directories: ["scripts", "texts", "results"]
 # haven't changed, it won't be re-run.
 workflows:
   setup:
+    - install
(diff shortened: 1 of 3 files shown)


benchmarks/speed/project.yml
```diff
@@ -43,13 +42,6 @@ assets:
 # via "spacy project run [command] [path]". The help message is optional and
 # shown when executing "spacy project run [optional command] [path] --help".
 commands:
-  - name: install
-    help: "Install dependencies"
-    script:
-      - "python -m pip install -r requirements.txt -f https://download.pytorch.org/whl/torch_stable.html"
-    deps:
-      - "requirements.txt"
-
```
svlandeg, line 52: What's the reasoning for removing this installation instruction from the pipeline? We have an `install` or `setup` action in most of the project workflows, and I think it makes it clear to the user what should happen before running anything else? We don't document anywhere that users should run `pip install -r` by default, right? And the  `-f` option has certainly helped me get a proper install for PyTorch before...

(sorry if I missed a related discussion)

svlandeg (commented):
Looks good, ran some tests on my machine. Just had a question on the `install` command.

benchmarks/speed/project.yml
adrianeboyd: This is related to #121 and the fact that this is relatively inconsistent across projects and hard to maintain in this form.

The `-f` URL doesn't work anymore for the newest versions (they modified how they structure their pip repos) and if I use it now in a new venv (with python 3.8 at least) I get a rocm wheel:

```
$ pip install torch -f https://download.pytorch.org/whl/torch_stable.html
Looking in links: https://download.pytorch.org/whl/torch_stable.html
Collecting torch
  Downloading https://download.pytorch.org/whl/rocm5.1.1/torch-1.12.0%2Brocm5.1.1-cp38-cp38-linux_x86_64.whl (1354.7 MB)
```

Some projects had `pip3`, some had `python -m pip`, some had `pip` and some had nothing. Plus people don't necessarily want to install the requirements with `pip` in the first place, or not by default because it can easily clobber things.

We can discuss whether these install steps make sense overall or if there's a more general way to make this easier for users.

svlandeg: Ok so would it make sense to keep the `install` command in just for this PR (we can still remove the `-f` flag) and address that bit in the other PR?

PR #121, opened 2022-07-20
Remove dependency installation from project configs

rmitsch:
## Goals

Remove dependency installation from project config files. 

The changes to install requirements per project in the scheduled BuildKitten job can be found in https://github.com/explosion/buildkite-pipelines/pull/9.

## Description

- Remove dependency installation steps/commands from project config files.
- Add removal of project dependencies in GH CI after test suite for project has finished.

### Types of change

Refactoring.

## Checklist


adrianeboyd:
Sofie probably wants to weigh in on this in general.

I'm not sure why the CI isn't triggered, though? Let me try closing and reopening...

run-tests.sh
```diff
@@ -14,14 +50,42 @@ do
...
       tested_dirs+=($full_second_level_dir)
       if [ -e $full_second_level_dir/requirements.txt ]; then
-        python -m pip install -r $full_second_level_dir/requirements.txt --extra-index-url https://download.pytorch.org/whl/cpu
+        $1 -m pip -q install -r $full_second_level_dir/requirements.txt --extra-index-url https://download.pytorch.org/whl/cpu --no-warn-script-location
+      fi
+
+      # Ensure proper spaCy version is installed.
+      spacy_version=$(
+        cat ${full_second_level_dir}/project.yml |
+        $1 -c "import yaml; import sys; print(yaml.safe_load(sys.stdin.read()).get('spacy_version', ''))"
```
adrianeboyd, line 66: I don't really understand why you don't just use `srsly` here? You can add it to the list of things that aren't uninstalled like `torch`?

rmitsch: Switched to `srsly`.

run-tests.sh
```diff
@@ -14,14 +50,41 @@ do
...
+
+      $1 -m pytest -q -s $full_second_level_dir
+
+      # Mark as failure if exit code isn't either 0 (success) or 5 (no tests found).
+      if [[ $? != @(0|5) ]]; then
+        status=1
+      fi
+
+      if [ -e $full_second_level_dir/requirements.txt ]; then
+        $1 -m pip freeze --exclude torch --exclude wheel cupy-cuda110 srsly aiohttp > installed.txt
```
adrianeboyd, line 80: This will be annoying to keep in sync, but:

```suggestion
        $1 -m pip freeze --exclude torch --exclude wheel cupy-cuda111 srsly aiohttp > installed.txt
```

run-tests.sh
```diff
@@ -14,14 +50,41 @@ do
...
   # Only run if change happened at second level, since first level-changes don't require the tests to be re-run.
+  # If change happened at first level, $second_level_dir will be empty.
   if [ ! -z "$second_level_dir" -a "$second_level_dir" != " " ]; then
     if [[ ! " ${tested_dirs[*]} " =~ " ${full_second_level_dir} " ]]; then
+      echo "Executing tests for $full_second_level_dir"
+
       tested_dirs+=($full_second_level_dir)
       if [ -e $full_second_level_dir/requirements.txt ]; then
-        python -m pip install -r $full_second_level_dir/requirements.txt --extra-index-url https://download.pytorch.org/whl/cpu
+        $1 -m pip -q install -r $full_second_level_dir/requirements.txt --extra-index-url https://download.pytorch.org/whl/cpu --no-warn-script-location
```
adrianeboyd, line 60: Can you assign `$1` to a variable with a clearer name so this is a bit easier to follow?

rmitsch:
So the Azure CI runs succeed, but the [BuildKite runs don't](https://buildkite.com/explosion-ai/spacy-projects/builds/77#0182251c-b49d-4f78-af86-afb7a9ee4e3d/664-1062). I think this is due to some version mismatches that are a consequence of not being able to properly uninstall all packages. So two possibilities here: 
- I try to patch this particular issue up and hope it was the last one.
- I set up an alternative version with a separate venv for each project.

I'm in favor of the latter, since it's easier to maintain. If the extra installation time is that horrid, we can still go back to the progressive uninstall/install way of setting the projects up. What do you think?

adrianeboyd:
The underlying reason is that I made a change in spacy v3.2.0 that broke the compatibility (which I shouldn't have), so if you get spacy v3.1 and `spacy-experimental` than you can get this error.

I think the solution is make sure `spacy-experimental` is uninstalled? Are you sure the install step is working correctly? You can see traces of it here:

https://buildkite.com/explosion-ai/spacy-projects/builds/77#0182251c-b49d-4f78-af86-afb7a9ee4e3d/664-1017

rmitsch:
> Are you sure the install step is working correctly? 

IMO the install step is working fine, it's the `pip uninstall` that's failing. Looking into it.

adrianeboyd:
I meant `uninstall`, sorry.

rmitsch:
> I meant `uninstall`, sorry.

Using a venv on BuildKite and a fix for the spaCy re-installation w.r.t. the project config files seems to [have resolved all issues](https://buildkite.com/explosion-ai/spacy-projects/builds/83#018225dd-f120-446b-a5cf-dfe29a71e68b).

adrianeboyd:
This is looking pretty good! I like that fact that it's now just one script to maintain.

run-tests.sh
```diff
@@ -14,14 +51,39 @@ do
...
+
+      $python_interpreter -m pytest -q -s $full_second_level_dir
+
+      # Mark as failure if exit code isn't either 0 (success) or 5 (no tests found).
+      if [[ $? != @(0|5) ]]; then
+        status=1
+      fi
+
+      if [ -e $full_second_level_dir/requirements.txt ]; then
+        $python_interpreter -m pip freeze --exclude torch --exclude wheel cupy-cuda111 srsly aiohttp > installed.txt
```
adrianeboyd, line 80: If it's always reinstalling the dev requirements below, then I think the excludes can be limited to just the large things that we don't want to wipe out:

```suggestion
        $python_interpreter -m pip freeze --exclude torch --exclude cupy-cuda111 > installed.txt
```

`pip freeze` also already automatically excludes `setuptools, pip, distribute, wheel` unless you specify `--all`.

svlandeg:
How is this going to affect users running existing projects end-to-end? They need to remember to install the requirements first?

adrianeboyd:
I don't think we've decided for sure what to do here. This shows how a good plan for how we could manage the requirements for our CI/tests to make sure that the requirements are also maintained, since having a bunch of them in `azure-pipelines.yml` wasn't a particularly good way to do it.

For users I could imagine:

* an automatic step in `spacy project run` that uses `importlib` to check whether the requirements are installed if `requirements.txt` is present and provides example install commands for users if not
* an automatic script that adds/maintains install steps within the projects that we maintain here
* ???

rmitsch:
> I don't think we've decided for sure what to do here. ...

I like the idea of automatically checking if the requirements are installed + providing install commands or even having users confirm is the requirements should be installed automatically. 
I think we could do that before running _any_ step though (shouldn't take too long to do so?), not as a separate step.

svlandeg:
> I think we could do that before running _any_ step though (shouldn't take too long to do so?), not as a separate step.

I'm not sure I fully understand what you mean, like if we do it before running other steps, it becomes a separate step, no? Don't mean to be pedantic, just confused 😛 

I definitely agree that this solution feels better for the CI/test suite of things, but I want to make sure things don't unnecessarily break for users. We don't really version the projects or anything, but workflows that checkout a given project and run some kind of `all` command will suddenly break if the requirements aren't installed anymore as part of the workflow. So at the very least we need to think about how to communicate this clearly.

Do we want something like `assets` so we'd have
```
spacy project install
spacy project assets
spacy project run ...
```
or such?

> 
>     * an automatic step in `spacy project run` that uses `importlib` to check whether the requirements are installed if `requirements.txt` is present and provides example install commands for users if not
> 

I like that idea too, but it might generate some unnecessary overhead if this is run on every single `spacy project run ...` command?

rmitsch:
> > I think we could do that before running _any_ step though (shouldn't take too long to do so?), not as a separate step.
> 
> I'm not sure I fully understand what you mean, like if we do it before running other steps, it becomes a separate step, no? Don't mean to be pedantic, just confused stuck_out_tongue

Yeah, I misunderstood what Adriane was saying - I meant exactly what she's saying in her first suggestion :smile: We could draft something and run some tests to see how much overhead it really is? I feel like this is the safest and lowest-effort (for users) solution.

If this doesn't work out for whatever reason, I'm in favor of `spacy project install`. Feels appropriate.

rmitsch:
I guess we want to try adding the requirements checking/automatic install (or, alternatively, `spacy project requirements/install`) to spaCy before we merge this PR then?

rmitsch:
Alright, so checking whether _just_ the packages in `requirements.txt` are installed seems to be not too much work and achievable with only `pkg_resources`. Getting the complete dependency tree is a bit more involved. 

What we'd be looking for here, ideally, is something like `pip check` to ensure all packages are installed with the right versions. This works with `conda` as well. Haven't tried `poetry` or other package managers yet. Small hiccup: installing spaCy from scratch in a new `conda` environment and then running `pip check` yields: 
```
thinc 8.0.15 has requirement pydantic!=1.8,!=1.8.1,<1.9.0,>=1.7.4, but you have pydantic 1.9.0.
spacy 3.3.1 has requirement pydantic!=1.8,!=1.8.1,<1.9.0,>=1.7.4, but you have pydantic 1.9.0.
```
So I guess we have to be somewhat lenient about version mismatches. In general I think the combination of 
- checking if all packages in `requirements.txt` are installed using `pkg_utils` and
- ensure dependency tree correctness/completeness with `pip check` 
might be a low-effort and relatively robust way, based only on standard tools, to check if everything is installed properly.

What do you think?

adrianeboyd:
I would only check for the packages in `requirements.txt` and ignore the full dependency check.

It doesn't look like the dependencies in the conda packages are incorrect (it should be `<1.9.0` for that version), so I'm not sure what's going on there.

rmitsch:
Suggestion for dependency checking: https://github.com/explosion/spaCy/pull/11226. This doesn't install dependencies yet, just detects. Only with an automatic install we'd be fully backwards compatible, but I feel like that's kind of intrusive for a default behavior...

benchmarks/nel/tests.py
```diff
@@ -1,16 +0,0 @@
-""" Testing all project steps. """
-
-from spacy.cli.project.run import project_run
-from spacy.cli.project.assets import project_assets
-from pathlib import Path
-
-
-def test_nel_emerson_project():
```
svlandeg, line 8: Why was this removed?

benchmarks/nel/test_project_nel_benchmark.py
```diff
@@ -0,0 +1,18 @@
+""" Testing all project steps. """
+
+from spacy.cli.project.run import project_run
+from spacy.cli.project.assets import project_assets
+from pathlib import Path
+
+
+def test_nel_benchmark_project():
```
svlandeg, line 8: You can probably remove this file from this PR, as it's present in https://github.com/explosion/projects/pull/115 as well?

experimental/ner_spancat_compare/project.yml
```diff
@@ -147,7 +147,6 @@ commands:
   - name: "install"
     help: "Install dependencies"
     script:
-      - "pip install --user -r requirements.txt"
       - "python -m spacy download en_core_web_lg"
```
svlandeg, line 150: Maybe this should be renamed to `download` + update the `help` bit.

integrations/fastapi/project.yml
```diff
@@ -1,5 +1,5 @@
 title: "FastAPI integration"
-description: "Use [FastAPI](https://fastapi.tiangolo.com/) to serve your spaCy models and host modern REST APIs. To install the dependencies and start the server, you can run `spacy project run start`. To explore the REST API interactively, navigate to `http://${vars.host}:${vars.port}/docs` in your browser. See the examples for how to query the API using Python or JavaScript."
+description: "Use [FastAPI](https://fastapi.tiangolo.com/) to serve your spaCy models and host modern REST APIs. To start the server, you can run `spacy project run start`. To explore the REST API interactively, navigate to `http://${vars.host}:${vars.port}/docs` in your browser. See the examples for how to query the API using Python or JavaScript."
```
svlandeg, line 2:
```suggestion
description: "Use [FastAPI](https://fastapi.tiangolo.com/) to serve your spaCy models and host modern REST APIs. To start the server, install the requirements first and then run `spacy project run start`. To explore the REST API interactively, navigate to `http://${vars.host}:${vars.port}/docs` in your browser. See the examples for how to query the API using Python or JavaScript."
```

tutorials/nel_emerson/project.yml
```diff
@@ -97,16 +96,9 @@ commands:
...
   # "run" list. They can depend on third-party libraries etc. Here are just
-  # some examples of what's possible.
-  - name: setup
-    help: Install dependencies
-    script:
-      - "python -m pip install -r requirements.txt"
-    deps:
-      - "requirements.txt"
-
+  # one example for what's possible.
```
svlandeg, line 101: This whole comment string can be removed - it seems to be a leftover from some template, there isn't even a "run" list in this project.

tutorials/ner_tweets/project.yml
```diff
@@ -124,9 +125,8 @@ workflows:
 # shown when executing "spacy project run [optional command] [path] --help".
 commands:
   - name: "install"
-    help: "Install dependencies"
-    script: 
-      - "pip3 install --user -r requirements.txt"
+    help: "Install models"
+    script:
```
svlandeg, line 129: I think we typically call this command `download` so maybe we stick with that for consistency?

svlandeg (commented):
Only had some more minor comments.

Either way we probably want to wait to merge this until https://github.com/explosion/spaCy/pull/11226 is merged and available in a next release of `spacy`.

benchmarks/nel/test_project_nel_benchmark.py
rmitsch: Yes, same as above - should have been in https://github.com/explosion/projects/pull/115 only. I'll revert.

rmitsch:
This should be restored to its original state before I skillfully erased all of it. I'll trigger the `all` tests on BuildKite - once these succeed, I'll update again asking for a quick once-over review.

rmitsch:
BuildKite run succeeds. A final review/check appreciated, but there's no hurry since we don't wanna merge this before spaCy 3.5 is released anyway.

integrations/wandb/requirements.txt
```diff
@@ -1,2 +1,2 @@
-ml_datasets>=0.2.0a0
+ml_datasets==0.2.0a0
```
adrianeboyd, line 1: Is this correct? (It seems unlikely that it only works with an alpha version?)

benchmarks/ud_benchmark/project.yml
```diff
@@ -1,7 +1,7 @@
...
 
-spacy_version: ">=3.2.1,<4.0.0"
+spacy_version: ">=3.3.0,<3.4.0"
```
adrianeboyd, line 4: Why is this change necessary?

benchmarks/textcat_architectures/requirements.txt
```diff
@@ -1 +1 @@
-ml_datasets>=0.2.0a0
\ No newline at end of file
+ml_datasets==0.2.0a0
```
adrianeboyd, line 1: Ditto about alpha release.

adrianeboyd (commented):
Can you rerun the tests with v3.4.2? It might also make sense to change the spacy requirements in the CI.

rmitsch:
> It might also make sense to change the spacy requirements in the CI.

To `>=3.4`?

rmitsch:
> Why is this change necessary?

Not doing that [results in](https://dev.azure.com/explosion-ai/Public/_build/results?buildId=21001&view=logs&j=f1d52934-b974-5ece-337e-e658a4411382&t=25745add-d8f1-5bdf-c7de-2e51e93603d8)
```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
spacy-experimental 0.5.0 requires spacy<3.4.0,>=3.3.0, but you have spacy 3.4.2 which is incompatible.
```

adrianeboyd:
Ah, then I would change it to:

```
spacy-experimental>=0.2.0,<0.7.0
```

azure-pipelines.yml
```diff
@@ -30,12 +30,12 @@ jobs:
                 versionSpec: '$(python.version)'
                 architecture: 'x64'
 
+          # Exact spaCy pin is temporary, will be removed
           - script: |
-                pip install "spacy>=3.2.0,<3.5.0"
+                pip install "spacy==3.4.2"
```
adrianeboyd, line 35: Since we don't get real interaction with the requirements check in the test suite (we're not raising warnings to errors), I think it's okay to just revert this to the previous `>=3.2.0,<3.5.0`.

tutorials/README.md
```diff
@@ -1,6 +1,6 @@
 <a href="https://explosion.ai"><img src="https://explosion.ai/assets/img/logo.svg" width="125" height="125" align="right" /></a>
 
-# 🪐 Project Templates: Tutorials (12)
+# 🪐 Project Templates: Tutorials (11)
```
adrianeboyd, line 3: This changes seems like it belongs in a separate PR, is there a reason for having it here (that I am missing in this huge set of changes)?

rmitsch: Not that I can recall - guess I updated the project summary docs when I updated the project docs to reflect changed step names. Should I revert these changes?

rmitsch: @adrianeboyd Is this ok as-is or should these changes be reverted?

rmitsch: Resolving as this was overwritten by changes in `master` anyway.

rmitsch:
@adrianeboyd @svlandeg This is ready to go IMO (a last sanity check might not hurt). The [dependency check](https://github.com/explosion/spaCy/pull/11226) was already included in [spaCy v3.4.3](https://github.com/explosion/spaCy/releases/tag/v3.4.3), so I think there's no need to wait for v3.5 to merge this.

azure-pipelines.yml
```diff
@@ -31,11 +31,10 @@ jobs:
                 architecture: 'x64'
 
           - script: |
-                pip install "spacy>=3.2.0,<3.5.0"
+                pip install "spacy>=3.2.0,<3.5.0"
```
adrianeboyd, line 34:
```suggestion
                pip install "spacy>=3.2.0,<3.5.0"
```

azure-pipelines.yml
```diff
@@ -31,11 +31,10 @@ jobs:
...
                 pip install pytest
                 pip install wheel
             displayName: 'Install dependencies'
 
-          - task: Bash@3
-            inputs:
-              filePath: 'run-tests.sh'
\ No newline at end of file
+          - script: bash run-tests.sh python
+            displayName: 'Run tests'
```
adrianeboyd, line 40: What is the difference in bash/azure here?

requirements.txt
```diff
@@ -0,0 +1,3 @@
+"spacy>=3.2.0,<3.5.0"
```
adrianeboyd, line 1:
```suggestion
spacy>=3.2.0,<3.5.0
```

adrianeboyd: Wait, this is the wrong place. The `spacy` requirement should be in `project.yml`.

adrianeboyd, line 1:
```suggestion
```

tutorials/README.md
```diff
@@ -9,11 +9,9 @@
...
 | [`ner_pytorch_medical`](ner_pytorch_medical) | Detecting entities in Medical Records with PyTorch |
 | [`ner_tweets`](ner_tweets) | Detecting people entities in tweets (Named Entity Recognition) |
-| [`parser_low_resource`](parser_low_resource) | Training a POS tagger and dependency parser for a low-resource language |
 | [`rel_component`](rel_component) | Example project of creating a novel nlp component to do relation extraction from scratch. |
-| [`spancat_food_ingredients`](spancat_food_ingredients) | Span Categorization in Prodigy |
 | [`textcat_docs_issues`](textcat_docs_issues) | Predicting whether a GitHub issue is about documentation (Text Classification) |
-| [`textcat_goemotions`](textcat_goemotions) | Categorization of emotions in Reddit posts (Text Classification) |
\ No newline at end of file
+| [`textcat_goemotions`](textcat_goemotions) | Categorization of emotions in Reddit posts (Text Classification) |
+| [`spancat_food_ingredients`](spancat_food_ingredients) | Span Categorization in [Prodigy](https://prodi.gy/) |
```
adrianeboyd: This needs to be updated again, double-check all the category READMEs again?

tutorials/ner_tweets/requirements.txt
```diff
@@ -1,10 +1,4 @@
-# NLP
-spacy>=3.1
```
adrianeboyd, line 2: This may also be a case where the spacy requirement should be added to `project.yml`?

azure-pipelines.yml
rmitsch: None that I noticed. The [Azure bash operator](https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/reference/bash-v3?view=azure-pipelines) is neater, but is just running `bash` under the hood. For our purpose just running `bash` directly seemed easier to read to me.

requirements.txt
rmitsch: I think this is valid. This is the fallback spaCy version. If the first project doesn't specify `spacy_version` in its `project.yml`, then we would have a python interpreter without `spacy` with the current version of `run-tests.sh`. If it does, it's overwritten anyway. There are other ways to ensure we have a valid `spacy` installation for the first project, but I think this is alright?

rmitsch: See https://github.com/explosion/projects/pull/121#discussion_r1066816818.

tutorials/ner_tweets/requirements.txt
rmitsch: It [has been](https://github.com/rmitsch/spacy-projects/blob/refactor/project-dependent-requirements/tutorials/ner_tweets/project.yml#L63). I don't remember why we bumped the version here - probably ran into some issues testing this with 3.1.

requirements.txt
adrianeboyd: Ah, I thought this was in a subdirectory, sorry. Then the CI step should install from `requirements.txt` rather than repeating this.

tutorials/ner_tweets/requirements.txt
adrianeboyd: I'd just remove it from `requirements.txt` here then.

tutorials/README.md
rmitsch: Ran `update_category_docs()`. Updated in https://github.com/explosion/projects/pull/121/commits/8e9fae061b5e3800f9bc430135b87784bba0b4ab.

adrianeboyd:
The `benchmarks/nel` project seems to have unrelated changes?

rmitsch:
> The benchmarks/nel project seems to have unrelated changes?

That's true, but these are just formatting. Do you mind leaving them in? It's no issue to revert those either.

benchmarks/nel/README.md
```diff
@@ -49,20 +47,9 @@ in the project directory.
 
 | File | Source | Description |
 | --- | --- | --- |
-| `assets/wiki/wikidata_entity_dump.json.bz2` | URL | Wikidata entity dump. Download can take a long time! |
-| `assets/wiki/wikipedia_dump.xml.bz2` | URL | Wikipedia dump. Download can take a long time! |
-| `assets/wiki/wikidata_entity_dump_filtered.json.bz2` | URL | Filtered Wikidata entity dump for demo purposes. |
-| `assets/wiki/wikipedia_dump_filtered.xml.bz2` | URL | Filtered Wikipedia dump for demo purposes. |
+| `assets/reddit.zip` | URL | Entity linking dataset scraped from Reddit. See [paper](https://arxiv.org/abs/2101.01228). |
```
adrianeboyd, line 50: What is the status of these edits? (This doesn't seem like it's just formatting?)

rmitsch: Ah, thanks for catching this. This is indeed not intentional.

rmitsch: Reverted non-formatting changes in https://github.com/explosion/projects/pull/121/commits/7de339bc19d82deda49f70014debabe252fed9a7.

benchmarks/nel/README.md
```diff
@@ -16,18 +16,16 @@ The following commands are defined by the project. They
 can be executed using [`spacy project run [name]`](https://spacy.io/api/cli#project-run).
 Commands are only re-run if their inputs have changed.
 
-| Command | Description |
```
adrianeboyd, line 19: None of the `benchmarks/nel/README.md` edits look like they belong in this PR?

rmitsch: Yeah, it seems like I forgot to update the readme properly in a prior PR. I regenerated the readme, it should match the project structure now.

rmitsch:
@adrianeboyd Do you feel the open issues have been addressed?

benchmarks/nel/README.md
adrianeboyd: Please merge in `v3` and move any remaining changes to `benchmarks/nel` that aren't related to the dependency installation to a separate PR.

adrianeboyd:
In this PR adding a correct `spacy_version` in `nel_emerson` would be okay, but if you're going to update the project code instead (which is fine!), please move these changes to a separate PR , and also add the correct `spacy_version` requirements.

rmitsch:
> In this PR adding a correct spacy_version in nel_emerson would be okay, but if you're going to update the project code instead (which is fine!), please move these changes to a separate PR, and also add the correct spacy_version requirements.

Was about to write a comment in this regard. In order for the `nel_emerson`  tests to succeed, we can either
- limit `spacy_version`to < 3.5 or
- update `KnowledgeBase` to `InMemoryLookupKB` and limit `spacy_version` to >= 3.5 or
- add qualified imports everywhere - something like:
```python
try:
    from spacy.kb import InMemoryLookupKB as KnowledgeBase
except ModuleNotFoundError:
    from spacy.kb import KnowledgeBase
```
Inclined towards the first option. What do you think?

adrianeboyd:
I realize that I don't think we highlighted how breaking the `KnowledgeBase` changes were in all the release notes.

For this PR, I think adding `spacy_version: ">3.0.0,<=3.5.0"` is fine. Any other options would be better as a separate PR (that would be merged before this one).

rmitsch:
> I realize that I don't think we highlighted how breaking the KnowledgeBase changes were in all the release notes.

Yeah, @svlandeg and me went over the issue of the KB changes being breaking a lot before merging it. I'm not super happy about this not being backward-compatible, but we didn't come up with a solution without drawbacks.

svlandeg:
> I realize that I don't think we highlighted how breaking the `KnowledgeBase` changes were in all the release notes.

You're right - should we go in and make that more explicit in the release notes?

adrianeboyd:
I think it would be good to add it explicitly to the release notes and usage docs, yes.

rmitsch:
Thanks for https://github.com/explosion/projects/pull/121/commits/3dc39b91f6d74aed85698277fd5f73e15b4ca884, @adrianeboyd! Was looking into it. Updated your change to reflect the new step name.

adrianeboyd:
Ah, bad merge on my part...

adrianeboyd:
(But I think that `capture=True` doesn't actually work correctly in general.)

adrianeboyd:
Yay, I think this is ready to merge!

PR #125, opened 2022-08-15
Track exit code in CI tests

adrianeboyd:


adrianeboyd:
I see now that this is addressed better in #121, but this will help with the currently open PRs for now.

PR #21, opened 2020-10-26
Experimental: relation extraction component

svlandeg:
Project template that implements a new pipeline component from scratch: a relation extraction component that labels pairs of entities with their appropriate relation(s). Used as example for the nightly/v3 docs: https://nightly.spacy.io/usage/layers-architectures#component-rel (**TODO**: update those docs after finalizing & merging this)

## Task definition
- Binary relation extraction: between exactly 2 spans, each span representing a relevant entity
- More than one label can be relevant for each entity pair (multi-label)
- Annotations are stored in `doc._.rel` and keyed by `(span1.start, span2.start)`. The values are `{label: prob}`, e.g. `{"BINDS": 0.4, "REGULATES": 0.9}`

## Model
Implemented in `scripts/rel_model.py`
* `tok2vec` component
   * produces `tokvecs` vectors for tokens
* `get_candidates` defines a list of `Span` pairs ("candidates") that we want to classify within a document
    * `rel_cand_generator.v2` imposes a (configurable) max distance and removes self-relations
* `create_candidate_tensor` is a function that transforms a list of documents and their `tokvecs` to a tensor representation of the candidates  
   *  `rel_cand_tensor.v1` takes the mean of each token's `tokvec` within the entity span, then `hstack`'s the first and second span of the candidate.
   * The `backprop` should probably be rewritten to be more succinct, but `scatter_add` scares me a little and I wanted to focus on it being correct/readable first 
* an output layer that determines a `[0,1]` value for each candidate and each label in the dataset 
    * `rel_output_layer.v1` currently uses `Linear >> Logistic`

## Pipeline component
Implemented in `scripts/rel_pipe.py`
* Implements the `TrainablePipe` skeleton, delegating the main parts to `self.model`
* Keeps track of the `labels` in the dataset
* Defines a `threshold` value to allow printing P/R/F scores on the dev set during training

## Data
I've been toying a bit with biomedical relations, but looking only at simple cases and only distinguishing between "regulates" and "binds". I've been doing this with Prodigy's `rel.manual` recipe and I intend on annotating a bit more soon. The data is transformed to `.spacy` files by `scripts/parse_data.py` and read back in by the custom `Gold_ents_Corpus.v1` which preserves both the gold tokenization and the gold entities (GGPs: gene or gene products). The texts are from the BioNLP ST'11 GE corpus, and originate from Pubmed & Pubmed Central.

## project.yaml
[...]

PR #128, opened 2022-08-30
Don't do pip user installs

polm:
Pip user installs go to a location that's not project specific and don't work in ordinary virtualenvs.

Additionally, since we normalize pip/pip3 in our projects run code, we don't need to use pip3 in scripts.


adrianeboyd:
I think this should be handled by #21 instead.

polm:
Ah, right, that's the right approach overall.

PR #129, opened 2022-09-12
NEL benchmark: Add comparison of multiple evaluation runs

rmitsch:
## Description

Adds a step for the comparison of multiple evaluations, to simplify comparing e.g. multiple candidate generators. This includes
- persisting evaluation results as `.csv` files for every execution of `spacy project run evaluate` in a new directory `evaluations` and
- a new workflow step merging all available individual evaluation results, sorting by run and then by model.

A run name can be configured (defaults to a stringified timestamp of format `%Y-%m-%d_%H-%M-%S`).

Example output:
```
============================ compare_evaluations ============================
Running command: env PYTHONPATH=. python ./scripts/compare_evaluations.py mewsli_9
16:30:45 INFO 
+-----------------------------------+-------------------+-----+----+----+---------+--------+-----------+
|                Run                |       Model       |  TP | FP | FN | F-score | Recall | Precision |
+-----------------------------------+-------------------+-----+----+----+---------+--------+-----------+
|        2022-09-12_16-07-40        |   Candidate Gen.  | 150 | 4  | 4  |  0.974  | 0.974  |   0.974   |
|        2022-09-12_16-07-40        |       Random      | 131 | 14 | 19 |  0.888  | 0.873  |   0.903   |
|        2022-09-12_16-07-40        |       Prior       | 145 | 0  | 5  |  0.983  | 0.967  |    1.0    |
|        2022-09-12_16-07-40        |       Oracle      | 145 | 0  | 5  |  0.983  | 0.967  |    1.0    |
|        2022-09-12_16-07-40        |    Context Only   | 145 | 5  | 5  |  0.967  | 0.967  |   0.967   |
|        2022-09-12_16-07-40        | Context And Prior | 145 | 5  | 5  |  0.967  | 0.967  |   0.967   |
| NEL benchmark_2022-09-12_15-21-03 |   Candidate Gen.  | 150 | 4  | 4  |  0.974  | 0.974  |   0.974   |
| NEL benchmark_2022-09-12_15-21-03 |       Random      | 128 | 17 | 22 |  0.868  | 0.853  |   0.883   |
| NEL benchmark_2022-09-12_15-21-03 |       Prior       | 145 | 0  | 5  |  0.983  | 0.967  |    1.0    |
| NEL benchmark_2022-09-12_15-21-03 |       Oracle      | 145 | 0  | 5  |  0.983  | 0.967  |    1.0    |
| NEL benchmark_2022-09-12_15-21-03 |    Context Only   | 145 | 5  | 5  |  0.967  | 0.967  |   0.967   |
| NEL benchmark_2022-09-12_15-21-03 | Context And Prior | 145 | 5  | 5  |  0.967  | 0.967  |   0.967   |
| NEL benchmark_2022-09-12_15-53-41 |   Candidate Gen.  | 150 | 4  | 4  |  0.974  | 0.974  |   0.974   |
| NEL benchmark_2022-09-12_15-53-41 |       Random      | 129 | 16 | 21 |  0.875  |  0.86  |    0.89   |
| NEL benchmark_2022-09-12_15-53-41 |       Prior       | 145 | 0  | 5  |  0.983  | 0.967  |    1.0    |
| NEL benchmark_2022-09-12_15-53-41 |       Oracle      | 145 | 0  | 5  |  0.983  | 0.967  |    1.0    |
| NEL benchmark_2022-09-12_15-53-41 |    Context Only   | 145 | 5  | 5  |  0.967  | 0.967  |   0.967   |
| NEL benchmark_2022-09-12_15-53-41 | Context And Prior | 145 | 5  | 5  |  0.967  | 0.967  |   0.967   |
| NEL_benchmark_2022-09-12_15-18-23 |   Candidate Gen.  | 150 | 4  | 4  |  0.974  | 0.974  |   0.974   |
| NEL_benchmark_2022-09-12_15-18-23 |       Random      | 130 | 15 | 20 |  0.881  | 0.867  |   0.897   |
| NEL_benchmark_2022-09-12_15-18-23 |       Prior       | 145 | 0  | 5  |  0.983  | 0.967  |    1.0    |
| NEL_benchmark_2022-09-12_15-18-23 |       Oracle      | 145 | 0  | 5  |  0.983  | 0.967  |    1.0    |
| NEL_benchmark_2022-09-12_15-18-23 |    Context Only   | 145 | 5  | 5  |  0.967  | 0.967  |   0.967   |
| NEL_benchmark_2022-09-12_15-18-23 | Context And Prior | 145 | 5  | 5  |  0.967  | 0.967  |   0.967   |
+-----------------------------------+-------------------+-----+----+----+---------+--------+-----------+
```

## Checklist


polm (commented):
I was able to run this and it looks OK overall. I had a few comments about places I think pathlib usage could be cleaned up but it's nitpicks. 

Not sure how sophisticated we want to make this, but if the goal is to find the best run, maybe:

1. have an export format that's more machine-readable, like a combined CSV?
2. highlight the run that maximizes some criteria, maybe just with an extra row that's empty or filled with `*`?

benchmarks/nel/scripts/datasets/dataset.py
```diff
@@ -393,11 +398,33 @@ def evaluate(
...
+    def compare_evaluations(self) -> None:
+        """Generate and display table for comparison of all available runs for this dataset."""
+        header: Optional[List[str, ...]] = None
+        lines: List[List[str, ...]] = []
+        dir_path = Path(os.path.abspath(__file__)).parent.parent.parent / "evaluation" / self.name
+
+        for path in [
+            dir_path / file_name for file_name in os.listdir(dir_path)
+            if os.path.isfile(dir_path / file_name) and file_name.endswith(".csv")
+        ]:
```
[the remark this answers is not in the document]
rmitsch: No, I agree - here readability matters more here than making sure it's fail-safe.

rmitsch:
> 1. have an export format that's more machine-readable, like a combined CSV?

You mean for the combined evaluation results table?

> 2. highlight the run that maximizes some criteria, maybe just with an extra row that's empty or filled with *?

Makes sense. I'll add an option to pick the criterion to highlight.

rmitsch:
Rows with highest scores are now logged in bold. Rows related to candidate generation are treated separately, since the other rows related to disambiguation. This is a bit confusing though, maybe we should move the candidate generation rows into a separate table?

Example output (here with precision as criterion for highlighting):
image

benchmarks/nel/project.yml
```diff
@@ -9,8 +9,10 @@ vars:
   gpu_id: ""
   use_filtered_dumps: "--use_filtered_dumps"
   training_max_steps: 1000
+  run: ""
+  eval_comp_highlight_metric: "F-score"  # one of ("F-score", "Recall", "Precision")
```
svlandeg, line 13: Maybe just `eval_highlight_metric`

svlandeg: Also, perhaps the options should just be "F", "p" and "r" to prevent typos? (note that one of the docstrings has "f1" instead of "F-score")

benchmarks/nel/scripts/datasets/dataset.py
```diff
@@ -246,13 +249,16 @@ def _load_resource(self, key: str, force: bool = False) -> None:
...
+        run_name: str,
         candidate_generation: bool = True,
         baseline: bool = True,
         context: bool = True,
         spacyfishing: bool = True,
         n_items: Optional[int] = None,
     ) -> None:
         """Evaluates trained pipeline on test set.
+        run_name (str): Run name.
+        highlight_criterion (str): Criterion to highlight in table. One of ("f1", "recall", "precision").
```
svlandeg, line 261: `highlight_criterion` isn't actually an input parameter of `evaluate`?

svlandeg (commented):
Looking good! Added some small suggestions.

polm:
>  > have an export format that's more machine-readable, like a combined CSV?
>
> You mean for the combined evaluation results table?

Yes, for the combined results. But without a definite use case it's not really important, so just and idea. 

I was able to run your changes locally without issue. The bold is also pretty subtle here, maybe you could do reverse video instead? For ANSI escapes I was going to suggest you use wasabi (which actually also has table formatting, now that I think of it), but it doesn't seem to support reverse video, though it'd be easy to add.

benchmarks/nel/scripts/datasets/dataset.py
```diff
@@ -393,11 +399,50 @@ def evaluate(
...
+                max_crit_non_cand_gen = max(max_crit_non_cand_gen, float(row[highlight_crit_idx]))
+            else:
+                max_crit_cand_gen = max(max_crit_cand_gen, float(row[highlight_crit_idx]))
+        for row in rows:
+            if (
+                row[1] != "Candidate Gen." and float(row[highlight_crit_idx]) >= max_crit_non_cand_gen or
+                row[1] == "Candidate Gen." and float(row[highlight_crit_idx]) >= max_crit_cand_gen
+            ):
+                for i in range(len(row)):
+                    row[i] = '\033[1m' + row[i] + '\033[0m'
```
polm, line 442: You can change it to reverse video like this, though since it seems to only affect the contents and the not entire row it doesn't look great. Underline (`4`) looks a bit better. Honestly it's also fine as-is though. 

```suggestion
                    row[i] = '\033[7m' + row[i] + '\033[0m'
```

rmitsch: Going with underline now, I agree it looks better. Thanks for the recommendation!

benchmarks/nel/scripts/datasets/dataset.py
rmitsch: C-P mistake, removed in next commit.

benchmarks/nel/project.yml
rmitsch: Renamed. Where did you see "f1"? Wasn't able to find it :thinking:

svlandeg: It's removed now :-) https://github.com/explosion/projects/pull/129/commits/0df2ca989a0409139e37bf1ab07bb8096b6518e3#diff-354e564d383eafab258d09191b03c1eb1d98735fafe655a01289db50c120d80dL261

rmitsch:
> Yes, for the combined results. But without a definite use case it's not really important, so just and idea.

Added an export, stored as `evaluation/{dataset name}/compare-{datetime.datetime.now().strftime('%Y-%m-%d_%H-%M-%S')}.csv`. Does this make sense to you?

Following up on the question in the description:
> Do we want to show the table by label (in analogy to the label-based table in spacy project run evaluate) as well?

polm:
> Added an export, stored as evaluation/{dataset name}/compare-{datetime.datetime.now().strftime('%Y-%m-%d_%H-%M-%S')}.csv. Does this make sense to you?

That's great, thank you! I was easily able to navigate/sort that table with a CSV tool. 

> Do we want to show the table by label (in analogy to the label-based table in spacy project run evaluate) as well?

Is this just about the formatting/appearance? I don't really have a strong opinion either way, so I think it's fine as-is.

rmitsch:
> Is this just about the formatting/appearance? I don't really have a strong opinion either way, so I think it's fine as-is.

No. If you run `evaluate`, you'll see two tables: one with the results per method (random/oracle/...), one with results per method _and_ label. In `compare_evaluation` there's only the former so far (so all results per method _and_ run), because I thought that listing all results for method x run x label would be too much info. Not sure about that though (especially since we have the export now).

polm:
Ah, I see what you mean about labels. I agree that having that all in the output for visual inspection feels like just too much info. Even considering that it could be in the machine-readable output, since we don't have a concrete need for it at the moment, I don't feel it's imperative to add it now.

benchmarks/nel/scripts/datasets/dataset.py
```diff
@@ -393,11 +400,58 @@ def evaluate(
...
 
         logger.info(dict(cand_gen_label_counts))
-        evaluation.EvaluationResults.report(tuple(eval_results))
+        evaluation.EvaluationResults.report(tuple(eval_results), run_name=run_name, dataset_name=self.name)
 
         self._nlp_best.config["incl_context"] = False
         self._nlp_best.config["incl_prior"] = False
 
+    def compare_evaluations(self, highlight_criterion: str) -> None:
+        """Generate and display table for comparison of all available runs for this dataset.
```
polm, line 409: Can you add a comment here about how candidate generation rows are treated separately? It is slightly confusing, but the data is similar enough I think it should be OK if there's an explanation somewhere.

polm (commented):
A few more small things, I think this is basically good though.

benchmarks/nel/scripts/datasets/dataset.py
rmitsch: Extended the description. Please resolve this conversation if you think it's ok (and feel free to approve the PR if you think nothing else is to be changed).

polm (approved):
Looks good! Thanks for all your work on this.

svlandeg (approved):
Thanks Raphael for the nice work, and Paul for the detailed review! Merging 🎉

PR #130, opened 2022-09-12
Allow ml_datasets 0.2.0

svlandeg:
Two example projects were pinning `ml_datasets` to exactly `0.2.0a0` even though we've had a `0.2.0` release since then.

### DIFF
--- a/benchmarks/textcat_architectures/requirements.txt
+++ b/benchmarks/textcat_architectures/requirements.txt
@@ -1,1 +1,1 @@
-ml_datasets==0.2.0a0
+ml_datasets>=0.2.0a0
--- a/integrations/wandb/requirements.txt
+++ b/integrations/wandb/requirements.txt
@@ -1,2 +1,2 @@
-ml_datasets==0.2.0a0
+ml_datasets>=0.2.0a0
 wandb>=0.12.4,<0.20.0


polm (approved):
Looks good! 

It [looks like](https://github.com/explosion/projects/search?p=1&q=0a0) several projects depend on `spacy-streamlist>=1.0.0a1` - that's fine as-is, but should we update it to no longer refer to the alpha version anyway?

svlandeg:
> It [looks like](https://github.com/explosion/projects/search?p=1&q=0a0) several projects depend on `spacy-streamlist>=1.0.0a1` - that's fine as-is, but should we update it to no longer refer to the alpha version anyway?

I'm not sure how that happened, as there is no release `1.0.0a1` for `spacy-streamlit` (there is a `0.2.0a0` for `ml-datasets` though). I can go ahead and update the streamlit ones too.

svlandeg:
> It [looks like](https://github.com/explosion/projects/search?p=1&q=0a0) several projects depend on `spacy-streamlist>=1.0.0a1` - that's fine as-is, but should we update it to no longer refer to the alpha version anyway?

I guess these were specified as such to include `v1.0.0rc1` and `v1.0.0rc2`?

polm:
OK, in that case it should be fine.

PR #132, opened 2022-09-16
Use pysqlite3 instead of sqlite3 to allow for support of upsert syntax in Python 3.7

rmitsch:
## Description

The weekly build currently fails on BuildKite to the usage of Python 3.7 and the `sqlite3` version shipped with it due to its lack of support of the SQLite upsert syntax used in the NEL benchmark project.
Since upgrading `sqlite3` isn't easily possible independent from the Python installation itself, this PR replaces `sqlite3` with the drop-in support `pysqlite3`.

## Checklist

### DIFF
--- a/benchmarks/nel/requirements.txt
+++ b/benchmarks/nel/requirements.txt
@@ -5,3 +5,4 @@ scikit-learn
 fuzzyset2
 spacyfishing
 virtualenv
+pysqlite3>=3.24.0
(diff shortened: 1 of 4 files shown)


svlandeg (approved):
Makes sense to me - I'll leave this unmerged for a bit in case Adriane wanted to comment as well.

PR #133, opened 2022-09-27
Add minimal spancat demo

ljvmiranda921:
## Description

cf. https://github.com/explosion/spaCy/discussions/9635, https://github.com/explosion/spaCy/discussions/11517, https://github.com/explosion/spaCy/discussions/11518

This PR adds a minimal `spancat_demo` project that demonstrates how to convert and train a `spancat` pipeline. It borrows heavily from the `ner_demo` project (the dummy data, command organization, etc.), with minimal changes (conversion script, config, etc.)

## Types of change

Enhancement

## Checklist

- [x]  I confirm that I have the right to submit this contribution under the project's MIT license.
- [x]  I ran the tests, and all new and existing tests passed.

### DIFF
--- a/pipelines/README.md
+++ b/pipelines/README.md
@@ -1,6 +1,6 @@
 <a href="https://explosion.ai"><img src="https://explosion.ai/assets/img/logo.svg" width="125" height="125" align="right" /></a>
 
-# 🪐 Project Templates: Pipelines (15)
+# 🪐 Project Templates: Pipelines (16)
 
 | Template | Description |
 | --- | --- |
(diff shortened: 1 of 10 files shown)


ljvmiranda921:
I'll take a closer look. I agree it's a bit weird, it's just two Examples though.

ljvmiranda921:
@adrianeboyd this doesn't answer our question but in the meantime, I increased `training.patience` to prevent early stopping. It seems that `spancat` takes more time to "ramp-up" than NER. Running with this change should reach 100% now.

ljvmiranda921:
For the sake of completeness, I reran this on (1) three different seeds (`[0, 42, 10010]`) and on a (2) GPU and CPU. For all cases it reached 100%. The only noticeable difference is that it learns faster on GPU, i.e., reaches 100% faster on a smaller number of steps. I've also updated the `project.yml` to make the seed configurable

On CPU, it does take a while to get to the accuracy we want. I also encountered long stretches of low recall, but on 1000-th step or so it starts going out of local optima again. 

I can also change the samples if need be. Aside from that, what else should I try? @adrianeboyd

adrianeboyd:
I can see examples where you get to 100% between 500-1000 steps, but I really think there must be a bug in the component or in the config if this does not converge faster, or really if this does not converge at all, which it doesn't seem to do. There are only 12 different tokens and the entities are identical between the train and dev examples. There is nothing about the entity tokens that the attributes do not capture unambiguously. I really think this may be pointing to a more serious issue?

ljvmiranda921:
Ah I suspect the same thing. I can probably start by checking the default config and play around it first, then proceed to figuring out what's happening in the component itself. Might take a bit long for this.

adrianeboyd:
The basics of the project are all fine here, it's just that I'd rather not publish it until we understand what's going on.

kadarakos:
I added a new training sample and with that I got 100%.

PR #135, opened 2022-09-30
NEL benchmark: Speed up evaluation

rmitsch:
## Description

Speed up evaluation in NEL benchmark by using `.prefer_gpu()` and `.pipe()`. Also: add `.gitignore` to `evaluation` directory.

### Types of change

Chore/enhancement.

## Checklist


benchmarks/nel/project.yml
```diff
@@ -120,13 +122,24 @@ commands:
       - "corpora/${vars.dataset}/train.spacy"
       - "corpora/${vars.dataset}/dev.spacy"
 
+  - name: infer
+    help: "Infer entities for test set."
+    script:
+      - "env PYTHONPATH=. python ./scripts/infer.py ${vars.dataset}"
+    outputs:
+      - "evaluation/${vars.dataset}/predictions.spacy"
```
polm, line 130: This should be `predicted_test_docs.spacy`, right?

polm (commented):
This runs fine for me and looks good besides the filename issue, fix that and it should be good to go.

rmitsch:
Training and inference currently still takes way too long (a few seconds per document) - this is due to the entitiy linker component, most likely related to the candidate generator (initialization). I'll look into that more and update this.

rmitsch:
After profiling: the average inference time for a single document with `.pipe()` is about 1 s, with the bulk of that going towards the candidate generator call. I think that's alright for the benchmark scenario.

I did some minor clean-up - would be nice if you have another once-over, but nothing of relevance should have been changed.

svlandeg:
Thanks, I'll have a look!

benchmarks/nel/scripts/candidate_generation/base.py
```diff
@@ -34,27 +29,23 @@ def __call__(
...
     @abc.abstractmethod
-    def _init_lookup_structure(
-        self, kb: KnowledgeBase, max_n_candidates: int, **kwargs
-    ) -> Any:
+    def _init_lookup_structure(self, kb: KnowledgeBase, max_n_candidates: int, **kwargs) -> Any:
         """Init container for lookups for new dataset. Doesn't do anything if initialized for this dataset already.
         kb (KnowledgeBase): KnowledgeBase containing all possible entity candidates.
         max_n_candidates (int): Max. number of candidates to generate.
+        similarity_threshold (float): Minimal similarity of candidates with mentions. All candidates with lower
+            similarity are ignored.
```
svlandeg, line 48: Why has this docstring been added?

rmitsch: Mistake, thanks for catching it. Removed in next commit.

benchmarks/nel/scripts/datasets/dataset.py
```diff
@@ -249,6 +247,33 @@ def _load_resource(self, key: str, force: bool = False) -> None:
             with open(self._paths["failed_entity_lookups"], "rb") as file:
                 self._failed_entity_lookups = pickle.load(file)
 
+    def infer_test_set(self):
+        """Infer entities for test set."""
+        self._load_resource("nlp_best")
+        spacy.prefer_gpu()
```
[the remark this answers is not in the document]
rmitsch: Just because of my ignorance :upside_down_face: Wasn't aware of that, fixed in next commit.

rmitsch:
New changes (sorry for requesting a review before everything was done, realized that too late!):
- `infer` step has been folded back into `evaluate` (since no GPU, but `pipe` with multiprocessing is able to handle this in a reasonable amount of time).
- `evaluate()` has been adjusted to only execute exactly one NEL variation (see updated issue description for more details).
- New config file for evaluation (this is purely optional, I figured it'd be more convenient than passing many arguments).

From my point of view this is done now, pending further reviews: the evaluation performance is sufficient, and all the things I've noticed to be bothersome during evaluation of Mewsli-9 have been rectified.

svlandeg (approved):
Looks good!

PR #136, opened 2022-10-10
Re-add evaluation of trained model + small fixes for metric tracking

rmitsch:
## Description

- During previous refactoring the evaluation of the trained model, the evaluation of the trained model was erroneously dropped. This is fixed here.
- Small improvements for metric tracking (number of candidates is considered in calculating false positives for candidate generation) and evaluation table formatting.
- Use `EntityLinker` v2 instead of v1.
- Reverts leftover testing code in `scripts.evaluate`.

### Types of change

Enhancement, fix.

## Checklist

### DIFF
--- a/benchmarks/nel/configs/nel.cfg
+++ b/benchmarks/nel/configs/nel.cfg
@@ -47,7 +47,7 @@ max_n_candidates = 50
 similarity_cutoff = 0.5
 
 [components.entity_linker.model]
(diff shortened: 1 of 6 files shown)


polm (approved):
Sorry this took me a while, I was thrown off a bit by upstream changes to mewsli. 

I spotted one minor thing that might be improved, but it's also fine to leave as-is. I was able to run the code without issue (after figuring out the weirdness from upstream changes).

PR #139, opened 2022-10-14
NEL benchmark: pin repo revision for Mewsli-9 source

rmitsch:
## Description

After a [recent upstream change in the `google-research` repo](https://github.com/google-research/google-research/commit/f324eb40813f0980553f934d6d07a6cd639f953c) setting up the venv for compiling the Mewsli-9 dataset didn't work anymore. This pins the SVN revision to sparsely checkout/`svn export` the most recent version before this changes.

### Types of change

Fix.

## Checklist


polm (approved):
Thanks for the quick fix! This ran for me in a clean env, which the `v3` branch currently does not.

PR #141, opened 2022-10-21
Update Transformer architecture in example projects

polm:
Several projects were using `TransformerModel.v1`, but the latest model is v3. This shouldn't affect any of the projects as written, but if a user modifies a config to make a pipeline of their own, they may run into bugs we fixed a while ago without realizing why. 

This just updates the architectures. Leaving in draft until everything has been run for testing (locally, in addition to auto tests).


polm:
Tested this with all the projects that don't require non-public data and everything worked as expected, so this should be fine.

PR #142, opened 2022-10-23
Accelerate WikiNER Pipeline with Speedster

diegofiori:
Following the conversation with @kadarakos, we have added a modification to the WikiNER pipeline to speed it up with [nebullvm](https://github.com/nebuly-ai/nebullvm).

Nebullvm is an open-source tool designed to accelerate AI inference of deep learning models in a few lines of code. Within the WikiNER pipeline, nebullvm optimizes BERT to achieve the maximum acceleration physically possible on the hardware used.

Nebullvm speeds up the WikiNER pipeline between 20% and 80%, with no impact on model accuracy. The library could further accelerate deep learning model inference by applying more aggressive optimization techniques, which may result in a slight change in model accuracy. For more information, refer to the [nebullvm documentation on GitHub](https://github.com/nebuly-ai/nebullvm).

Below are the response times of the WikiNER pipeline in milliseconds (ms).

| Hardware | Original latency [ms] | Nebullvm optimized latency [ms] | Nebullvm speed-up |
| --- | --- | --- | --- |
| **Intel** | 139 | 114 | 1.2x |
| **AMD** | 293 | 162 | 1.8x |
| **Nvidia RTX 3090Ti** | 24.1 | 14.1 | 1.7x |
| **M1 Pro** | 143 | 121 | 1.2x |

Nebullvm is a project maintained by Nebuly, an open-source AI community working on solutions to address AI infrastructure bottlenecks and reduce computational costs, complexity and get models into production faster.


adrianeboyd:
Thanks, this looks like a nice PR! Can you explain a bit about the numpy error in windows that's the reason for skipping the tests?

morgoth95:
> Thanks, this looks like a nice PR! Can you explain a bit about the numpy error in windows that's the reason for skipping the tests?

That error arises because of the line below from the original `ner_wikiner` pipeline. 
```
@pytest.mark.skip(reason="TODO: numpy error on Windows")
```
Here’s the link to the [original test](https://github.com/explosion/projects/blob/v3/pipelines/ner_wikiner/test_project_ner_wikiner.py).

adrianeboyd:
Ah, that explains that. That was long enough ago that I have no idea about the details, but I was able to remove it for the original project and the tests passed with no problems (#145).

I think for this project you'd need to add a `requirements.txt` with the additional requirements, and I'm not sure how much of the project it would be feasible to run in the CI at all? The test suite sets very low defaults for common training setting like `max_steps` and `max_epochs` just to test that projects run as intended without spending a lot of time on actual training, and you can also modify the test to run individual steps rather than `"all"` or to modify overrides as makes sense, e.g.:

https://github.com/explosion/projects/blob/v3/tutorials/ner_multiple_trials/test_ner_multiple_trials.py

morgoth95:
Hi @adrianeboyd, I have updated the tests. We ran the test for Linux and it works smoothly. Note that the current version of Nebullvm does not yet support Windows devices. We are working on it and we will release a Windows compatible version early next year. As you will see, we are skipping testing for Windows, but if you have any suggestions I am happy to hear them!

experimental/ner_wikiner_accelerated_with_nebullvm/test_project_ner_wikiner.py
```diff
@@ -0,0 +1,15 @@
...
+
+from spacy.cli.project.run import project_run
+from spacy.cli.project.assets import project_assets
+
+
+def test_wikiner_project():
+    if sys.platform.startswith('win'):
+        # skip test on windows since nebullvm is not supporting windows
+        #  devices yet. 
+        return
```
adrianeboyd, line 12:
```suggestion
import pytest

from spacy.cli.project.run import project_run
from spacy.cli.project.assets import project_assets

@pytest.mark.skipif(sys.platform == "win32", reason="nebullvm does support windows devices yet")
def test_wikiner_project():
```

experimental/ner_wikiner_accelerated_with_nebullvm/scripts/extra_components.py
```diff
@@ -0,0 +1,265 @@
...
+    #     nebullvm_path = Path(path).parent / "nebullvm"
+    #     if nebullvm_path.exists():
+    #         print("Nebullvm loaded")
+    #         optimized_model = LearnerMetadata.read(path).load_model(path)
+    #         # warmup
+    #         _ = optimized_model(*optimized_model.get_inputs_example())
+    #         self._nebullvm_layer = copy.deepcopy(self.layers[0])
+    #         self._nebullvm_layer.shims[0]._hfmodel.transformer = optimized_model
+    #     return self
+
```
adrianeboyd, line 163:
```suggestion
```

experimental/ner_wikiner_accelerated_with_nebullvm/configs/default.cfg
```diff
@@ -0,0 +1,134 @@
...
+init_tok2vec = null
+vectors = null
+
+[system]
+gpu_allocator = null
+seed = 0
+
+[nlp]
+lang = "en"
+pipeline = ["tok2vec","ner"]
```
adrianeboyd, line 14: It may be confusing to users that the `transformer` component is named `tok2vec`, since there's a separate `tok2vec` component. (Yes, the naming here is confusing on our end.)

experimental/ner_wikiner_accelerated_with_nebullvm/scripts/compairison_script.py
```diff
@@ -0,0 +1,45 @@
+import time
```
adrianeboyd, line 1: Rename to `comparison_script.py`?

experimental/ner_wikiner_accelerated_with_nebullvm/scripts/extra_components.py
```diff
@@ -0,0 +1,265 @@
...
+            tokenizer=tokenizer,
+            store_latencies=True,
+            ignore_compilers=["tensor RT"],
+            tokenizer_args=dict(
+                add_special_tokens=True,
+                return_attention_mask=True,
+                # return_offsets_mapping=isinstance(
+                #     tokenizer,
+                #     PreTrainedTokenizerFast
+                # ),
```
adrianeboyd, line 82: Just out of curiosity: were there bugs related to this option?

experimental/ner_wikiner_accelerated_with_nebullvm/requirements.txt
```diff
@@ -0,0 +1,3 @@
+nebullvm>=0.4.4,<0.5.0
+spacy-transformers
+thinc
```
adrianeboyd, line 3:
```suggestion
spacy-transformers>=1.1.0,<1.2.0
```

adrianeboyd (commented):
The tests look good, just a few minor comments!

morgoth95:
Hello @adrianeboyd! I implemented the changes you required on both test and configuration files. Let me know if something else is needed for this PR. Cheers

adrianeboyd:
Only one more question: what parameters can you change or add for this command so that it runs faster for the test?

```
python scripts/load_optimize_and_save.py -d corpus/dev.spacy -m training/model-best/transformer/model -ot unconstrained -at 0.1
```

morgoth95:
> Only one more question: what parameters can you change or add for this command so that it runs faster for the test?
> 
> ```
> python scripts/load_optimize_and_save.py -d corpus/dev.spacy -m training/model-best/transformer/model -ot unconstrained -at 0.1
> ```

We can remove the `-at` argument and replace `-ot unconstrained` with `-ot constrained`.

adrianeboyd:
Thanks for the help!

When testing locally, I was surprised by the auto-installing from within the tool (which can make a lot of unexpected modifications to your venv) and also by the telemetry, so I think I'd want to add notes/warnings about this to the README.

I'm also not entirely sure I understand how/where/why everything is autoinstalled, and I'd like to disable the telemetry for the tests, but I think this is only possible if some dependencies are installed from `requirements.txt`. I'm not sure whether it's a mistake to always install `openvino` in all cases in order to turn all the telemetry off by default?

adrianeboyd:
I was also running a bit more testing with your generic instructions locally and I think you should try to avoid using the package `nvidia-pyindex`. I think you could add additional pip install options like `--no-cache-dir` or `--extra-index-url` to the specific package installs instead?

The problem is that `nvidia-pyindex` writes every single possible pip config file on install and sets `no_cache_dir = true` globally for any use of pip. In addition, the pip config changes aren't removed if you uninstall the package, so the only way to revert this is to manually delete or edit the files. I don't think packages should ever try to edit your pip configs, but even if they do, turning off the cache entirely is extremely bad behavior.

adrianeboyd:
I was working on writing a warning for the README but I see that you've resolved some of the issues upstream already. Should this wait until the next release of nebullvm?

diegofiori:
Hello @adrianeboyd. Sorry for the delay in responding. As you noticed, we have been working on fixing some of the problems you highlighted. I plan to refactor the code a bit to fit the new version of nebullvm.

diegofiori:
Hello @adrianeboyd! I finished the code refactor in order to support our latest release. We have massively refactor `nebullvm` transforming it into a framework for building optimization Apps/APIs. We then created a new library, `speedster`, for optimizing with a single line of code DL-models on target devices. In other words, the old nebullvm API has been moved into the `speedster` library. 
Let me summarise the main points of the new release:
* The new library does not install all compilers during the first import, but each compiler can be selected by the module and installed using our auto-installer module
* We fixed the problem you pointed out about the nvidia-pyindex package.

experimental/ner_wikiner_speedster/requirements.txt
```diff
@@ -0,0 +1,4 @@
+nebullvm>=0.6
+speedster>=0.0.1
```
svlandeg, line 2: Question: do we still need to specify `nebullvm` in the requirements, or is it going to be pulled by `speedster anyway?

svlandeg:
I'm a little confused why the test on Windows succeeded earlier today but not in the last run 🤔

pipelines/ner_wikiner/project.yml
```diff
@@ -31,7 +31,7 @@ commands:
     # caching works correctly.
     script:
       - "python scripts/partition.py assets/aij-wikiner-en-wp2.bz2 assets/iob"
-      - "python -m spacy convert assets/iob corpus --n-sents 10"
+      - "python -m spacy convert assets/iob corpus --converter iob --n-sents 10"
```
adrianeboyd, line 34: I think the bug related to this was fixed in `master` and this change isn't necessary?

adrianeboyd, line 34:
```suggestion
      - "python -m spacy convert assets/iob corpus --n-sents 10"
```

experimental/ner_wikiner_speedster/project.yml
```diff
@@ -0,0 +1,71 @@
...
+    - evaluate
+
+commands:
+  - name: corpus
+    help: "Convert the data to spaCy's format"
+    # Make sure we specify the branch in the command string, so that the
+    # caching works correctly.
+    script:
+      - "python scripts/partition.py assets/aij-wikiner-en-wp2.bz2 assets/iob"
+      - "python -m spacy convert assets/iob corpus --converter iob --n-sents 10"
```
adrianeboyd, line 36:
```suggestion
      - "python -m spacy convert assets/iob corpus --n-sents 10"
```

experimental/ner_wikiner_speedster/requirements.txt
```diff
@@ -0,0 +1,4 @@
...
+openvino-dev[onnx]
```
adrianeboyd, line 3: I also only added this to try to turn off the telemetry in the tests, but I'm not sure whether it needs to be installed separately or if there's a better way to manage this. (Maybe it's not installed by default anymore for the tests?)

experimental/ner_wikiner_speedster/test_project_ner_wikiner_speedster.py
```diff
@@ -0,0 +1,17 @@
+from pathlib import Path
+
+from spacy.cli.project.run import project_run
+from spacy.cli.project.assets import project_assets
+
+
+def test_wikiner_speedster_project():
+    import openvino_telemetry as tm
+    tm.Telemetry.opt_out("UA-17808594-29")
```
adrianeboyd, line 9: If `openvino` is removed from the requirements, then also remove this.

adrianeboyd:
Everything looks good, thanks again for the PR!

pipelines/ner_wikiner/project.yml
svlandeg: I just felt like it wouldn't hurt to be specific here. But yes, in theory the bug should be fixed on `master` as well!

PR #150, opened 2022-12-08
Add creation of projects.jsonl in docs update script

polm:
This adds the creation of a projects.jsonl file to the docs update script. The resulting file can be used to get a list of projects in the repo, which can be useful for an interactive project selector, for example.


.github/update_docs.py
```diff
@@ -14,11 +15,25 @@ def main(root: Path = typer.Argument(Path.cwd(), help="Root path to look in")):
...
+        entry = {"shortname": f"{path.parent.name}/{path.name}"}
+        entry["title"] = config["title"]
+        entry["description"] = config.get("description", "")
+        entries.append(entry)
+
+
+    with open("projects.jsonl", "w") as jsonfile:
+        for entry in entries:
+            jsonfile.write(json.dumps(entry))
+            jsonfile.write("\n")
```
svlandeg, line 35: It would be nice to print a message to the command line that acknowledges this has happened, similar to how the rest of the command works

svlandeg (commented):
I can totally see how this can be useful, thanks!

One thing we need to consider is that you can run `update_docs.py` also for a subdirectory. Do we want to have these `jsonl` files also in `benchmarks`, `experimental` etc? And how do we prevent this being generated if `update_docs.py` is run on a single project? It would currently generate an empty .jsonl which is clearly not ideal.

projects.jsonl
```diff
@@ -0,0 +1,46 @@
...
...--------|-----------|\n| DNA       |    0.70     | 0.36     |  0.47     |\n| protein   |    0.77     | 0.52     |  0.62     |\n| cell_line |    0.77     | 0.30     |  0.44     |\n| cell_type |    0.76     | 0.62     |  0.68     |\n| RNA       |    0.77     | 0.25     |  0.38     |\n| **Overall**| 0.76       | 0.47     |  0.58     |\n\n#### NER Results\n\nNER performs well against SpanCat for all entity types, but note that this\nprocess entails training five (5) models per entity type. This might work if\nyou have a small number of entities, but can be computationally heavy if you\nhave a lot.\n\n|           |   Precision | Recall |   F-score |\n|-----------|-------------|--------|-----------|\n| DNA       |    0.74     | 0.63   |  0.68     |\n| protein   |    0.76     | 0.72   |  0.74     |\n| cell_line |    0.74     | 0.57   |  0.64     |\n| cell_type |    0.78     | 0.72   |  0.75     |\n| RNA       |    0.86     | 0.65   |  0.74     |\n\nSince we have five (5) separate models in NER, what we can do afterwards is\ncombine them into a single `Doc` that transfers `doc.ents` to `doc.spans`. Since\nthe tokens are the same, we don't need to worry about misalignments and the like.\n"}
```
svlandeg, line 17: It seems to me that some of the "description" fields have been slightly "abused" to add all kinds of additional information that should only go in the readme. We should clean those out so that the resulting jsonl here is much more compact & clear.

polm:
I hadn't considered the case where the script was run on subdirectories. It only makes sense to have this json at the top level of the repo, so I just split the functionality into a separate script.

projects.jsonl
polm: I think it's OK for the description to be long and README-like, though the inconsistency in the way we do this is definitely an issue. For now I'll modify a couple of READMEs to have less in the description and more in the non-automated part and see how that looks.

polm: For all of the project files where there was a lengthy description, I moved most of it to the actual README file. There are a few borderline ones, like `pipelines/polar_component`, where I just left them as-is, but they could be split up.

Doing this I can see why everything ends up in the project file - the auto-generated docs contain the title, which is a natural start, and end with a list of steps etc., which is nice to have but more supplementary. So in between those is a natural place to put docs, but it's not possible to do that with the auto generated docs unless everything goes in the project file. 

I think things are still OK with the changes here, but maybe something to think about for the future.

polm:
I modified all very long descriptions in project files so that they are in README files instead. There are a few that were kind of borderline (like `pipelines/polar_component`) that I left alone. So things should be consistent now.

One thing that came up modifying this is that it became clear why some descriptions got long: the auto-docs include the title (which should be first) and also autogenerated tables of commands etc., which are nice to have but less important. So in the middle feels like a natural place to put docs. 

I think this edit is OK anyway, but something to keep in mind I guess.

.github/update_projects_json.py
```diff
@@ -0,0 +1,35 @@
...
+    msg.info(f"Updating projects.json in {root}")
+    entries = []
+    # We look specifically for project directories
+    for path in root.glob(f"**/*/{PROJECT_FILE}"):
+        path = path.parent
+
+        # prep data for the json file
+        config = load_project_config(path)
+        entry = {"shortname": f"{path.parent.name}/{path.name}"}
+        entry["title"] = config["title"]
```
rmitsch, line 24: Can we guarantee that `title` will always be available?

rmitsch (approved):
Looks reasonable to me. Maybe at some point we should run the update scripts automatically in the CI?

.github/update_projects_json.py
polm: In the schema the title is technically optional, so in general, we can't guarantee it's there. However, for this specific repo, I would consider it an error for any project to not have a title.

polm:
I think it'd make sense to have the data building scripts run regularly, like we do for black in spaCy.

rmitsch:
> I think it'd make sense to have the data building scripts run regularly, like we do for black in spaCy.

So including it in the `pre-commit` config?

rmitsch (approved):

polm:
Thanks for the reviews. On one last look-over I noticed some of the READMEs that should have had content moved had it deleted instead, so I fixed that. 

Since this is otherwise OK I'll go ahead and merge it.

PR #151, opened 2022-12-12
Add SpanRuler project

ljvmiranda921:
## Description

This example project demonstrates how one can use the [SpanRuler](https://spacy.io/api/spanruler) component for rule-based named entity recognition (NER). In spaCy v3 and below, this functionality can be achieved via the [EntityRuler](https://spacy.io/api/entityruler). However, we will start deprecating the `entity_ruler` component in v4 in favor of `span_ruler`.

I plan to create a separate project for SpanRuler using a dataset that's more fitted to span labelling. 

## Types of change

New project

## Checklist

- [x]  I confirm that I have the right to submit this contribution under the project's MIT license.


tutorials/spanruler_restaurant_reviews/scripts/attach_rules.py
```diff
@@ -0,0 +1,49 @@
...
+
+DEFAULT_PATTERNS_FILE = Path("patterns.jsonl")
+
+
+class Location(str, Enum):
+    before = "before"
+    after = "after"
+
+
+def attach_rules(
```
kadarakos, line 20: I'm not sure if its a better approach, but I would prefer having a `config.cfg` for the full pipeline instead of running the `attach` script. 

This can be achieved I think by just implementing a factory

```python
def build_restaurant_ruler(
    nlp: Language,
    name: str,
    rules: Dict[str, Dict[str, str]],
    *,
    fields: Optional[List[str]] = None
) -> SpanRuler:
)
```
If you follow my suggestions with the rules being in a version function than in the config you can add something like:


```yaml
[components.restaurant_ruler]
factory = "RestaurantRuler.v1"
rules = {"@misc" : "restaurant_rules.v1"}
```

This way there is no script to stitch the `nlp` together, rather, we have registered factories for `SpanRuler` and its rules which are versioned and there is an additional config that has the combined component. I think this is more explicit and easier to version in the future. Maybe no so important, but I like this design.

kadarakos (commented):
I only have some minor observations on wording and suggesting a different design for versioning the rules and constructing the combine pipeline.

I also have a question. Why does it matter if the `span_ruler` is strictly before the `ner`? Maybe I misunderstand, but it seems to me that it can come after the `ner` and take as `ents_filter` the `prioritize_new_ents_filter`?

tutorials/spanruler_restaurant_reviews/patterns.jsonl
```diff
@@ -0,0 +1,35 @@
+{"label": "Rating", "pattern": [{"IS_DIGIT": true}, {"LOWER": "star"}, {"LOWER": "review", "OP": "?"}]}
```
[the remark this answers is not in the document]
ljvmiranda921: For this I don't mind this approach either. I used the `patterns.jsonl` file because it's the most apparent one in [the documentation](https://spacy.io/usage/rule-based-matching/#spanruler-patterns). Using a registry makes sense too, but I haven't seen this one documented somewhere (I may be wrong) and it might cause a confusion?

tutorials/spanruler_restaurant_reviews/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+|          | NER only  | With Spanruler  |
+|----------|-----------|-----------------|
+| Price    | 76.67     | **77.99**       |
+| Rating   | 72.22     | **75.00**       |
+| Hours    | 66.67     | **67.11**       |
+| Amenity  | 65.15     | **65.42**       |
+| Location | **82.41** | 82.31           |
+
+Overall, we have better performance for the combined `ner` and `span_ruler`
```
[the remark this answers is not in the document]
ljvmiranda921: Oh there's an extra `l` in `overall`. I can edit this manually.

ljvmiranda921:
> I also have a question. Why does it matter if the span_ruler is strictly before the ner? Maybe I misunderstand, but it seems to me that it can come after the ner and take as ents_filter the prioritize_new_ents_filter?

@kadarakos Ah that's interesting. Is there a difference between the two? (Putting it before vs. putting it after then adding a filter that prioritizes the new entities?). The reason I put it before is because when there are overlaps, I want to prioritize the entities from the span ruler. 🤔

tutorials/spanruler_restaurant_reviews/patterns.jsonl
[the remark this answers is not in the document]
kadarakos: hmm true its not documented. I prefer it due to the versioning, wonder what others think also about this pattern vs. the `jsonl`. I think for packaging it also makes more sense to have the rules in the code?

tutorials/spanruler_restaurant_reviews/README.md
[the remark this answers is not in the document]
kadarakos: sorry updated the comment.

tutorials/spanruler_restaurant_reviews/scripts/attach_rules.py
adrianeboyd: I would version the rules but not the factory. `.` is no longer supported in factory names either, factories are intended to be versioned by the installed module version, so you can continue to have `nlp.add_pipe("ner")` across multiple versions of spacy and not have to remember some specific version. But spacy v3.1.x always gets you the exact same component, while there might be slight changes across minor versions otherwise.

adrianeboyd:
The `ner` component will make (slightly) different predictions depending on the existing annotation so choosing before/after can make a difference in the output.

kadarakos:
> The reason I put it before is because when there are overlaps, I want to prioritize the entities from the span ruler.

This is why I wanted to bring it up the prioritize functions. I was thinking it might be nice to show them off in this `project`. My understanding is that rather than considering the order of the components the `span_ruler` can always come on top and either take [`prioritize_new_ents_filter`](https://github.com/explosion/spaCy/blob/master/spacy/pipeline/span_ruler.py#L115) as `ents_filter` or [`prioritize_existing_ents_filter`](https://github.com/explosion/spaCy/blob/master/spacy/pipeline/span_ruler.py#L146). If `span_ruler` is after the names of these functions make it explicit if you'd like to prioritize `ner` or `span_ruler`.

tutorials/spanruler_restaurant_reviews/patterns.jsonl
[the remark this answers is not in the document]
ljvmiranda921: Yeah I like the idea of versioning rules too! Makes for a great "best practice" thing if possible. We can also update the docs if everyone agrees.

tutorials/spanruler_restaurant_reviews/scripts/attach_rules.py
kadarakos: Oh true, the rules are versioned and the factory takes the rules so there is no need to version the factory in any case. And what do you think about having the factory and adding the `span_ruler` in the config vs. in the `attach` script? To me the factory + config is the better more `spaCy`-like practice due to the explicitness.

tutorials/spanruler_restaurant_reviews/configs/spanruler.cfg
```diff
@@ -0,0 +1,154 @@
...
+batch_size = 1000
+disabled = []
+before_creation = null
+after_creation = null
+after_pipeline_creation = null
+tokenizer = {"@tokenizers":"spacy.Tokenizer.v1"}
+
+[components]
+
+[components.span_ruler]
```
ljvmiranda921, line 23: @kadarakos I did what you suggested. I haven't tested this yet, but you might want to take a look if the set up is correct.

tutorials/spanruler_restaurant_reviews/configs/spanruler.cfg
```diff
@@ -0,0 +1,154 @@
...
+vectors = ${paths.vectors}
+init_tok2vec = ${paths.init_tok2vec}
+vocab_data = null
+lookups = null
+before_init = null
+after_init = null
+
+[initialize.components]
+
+[initialize.components.span_ruler]
```
ljvmiranda921, line 151: @kadarakos The list of rules are now defined here as a registered function

tutorials/spanruler_restaurant_reviews/configs/spanruler.cfg
ljvmiranda921: I decided to create two separate configs because I want to compare a NER only pipeline and one with SpanRuler. Is this the best way to do it (i.e., two separate configs?). cc: @kadarakos

tutorials/spanruler_restaurant_reviews/scripts/attach_rules.py
ljvmiranda921: Sure will do that and remove the attach script to make everything based from the config

ljvmiranda921:
@kadarakos after poking around, I find that the configuration overall is elegant, but it's not fast and iterative. Unless I'm using it incorrectly: 
- If I want to modify the rules, I will update the rules file,  then run `spacy train` to train a model with those rules. Doing that takes a lot of time. 
- If I were to attach the rules, I can rerun the `attach` script over and over again while the baseline model stays the same. I only need to run `spacy train` once. 
 
Are there any optimizations you have in mind? If none, I can keep the `attach` script then mention that during iteration they can use X, the in the final production pipeline they should use Y.

kadarakos:
> Are there any optimizations you have in mind? If none, I can keep the `attach` script then mention that during iteration they can use X, then in the final production pipeline they should use Y.

I think the simplest and to me the nicest `spaCy` way is to have a `ner.cfg` that is a config for training the `ner` component and have a `ner_ruler.cfg` that uses `source` for the trained `ner` component and adds the `span_ruler` on top. Then `spacy assemble` can be used in place of `attach` to put the `span_ruler` on top of the trained `ner` and save the combined pipeline.

kadarakos:
Pushed a version that adds a new config `ner_ruler.cfg` which uses `source` to initialize the `ner` component from the trained model and in `[components]` it only adds the `span_ruler` as a new pipe. Unfortunately what I observed is **lower performance after adding `span_ruler` on top**. Please feel free to roll back my changes and also sorry I'm not sure why the results are worse. But I hope at least I managed to show what I mean by train `ner` once and then add the rule-based component on top of it with `assemble`. I think this is the right pattern, but there must be something no quite right in there if the results are lower.

ljvmiranda921:
I was actually thinking that it's nice to do the assemble thing. Hmm, not sure why the results are lower. I'll investigate tomorrow for a bit then if not I can roll-back!

kadarakos:
> I was actually thinking that it's nice to do the assemble approach. Hmm, not sure why the results are lower. I'll investigate tomorrow for a bit then if not I can roll-back!

I also think its nicer, it would be better to try to make it work somehow.

ljvmiranda921:
Hmm, I rerun things again and it seems that `span_ruler` is higher. But this time not so much. I'll push my commits and then you can replicate as well. 
image

ljvmiranda921:
I'm having a bit of a trouble understanding this error message:

```
E           subprocess.SubprocessError: Error running command:
E           
E           /opt/hostedtoolcache/Python/3.7.15/x64/bin/python -m spacy assemble configs/ner_ruler.cfg models/ner_ruler --components.tok2vec.source training/ner/model-best/ --components.ner.source training/ner/model-best/ --code scripts/rules.py
E           
E           Subprocess exited with status 1
E           
E           Process log (stdout and stderr):
E           
E           ✘ Error parsing config overrides
E           training -> max_epochs	not a section value that can be overwritten
```

Running `spacy project run . --vars.gpu_id 0` from my machine works well though. 🤔 I'll investigate a bit later.

ljvmiranda921:
Ah I get it now. It's because of [`SPACY_CONFIG_OVERRIDES`](https://github.com/explosion/projects/blob/611c5ddadbbb1f93c775e91d9aae2ac639edebd2/azure-pipelines.yml#L10). The `max_epochs` and `max_steps` params should also be included in `ner_ruler.cfg`

kadarakos:
> Ah I get it now. It's because of [`SPACY_CONFIG_OVERRIDES`](https://github.com/explosion/projects/blob/611c5ddadbbb1f93c775e91d9aae2ac639edebd2/azure-pipelines.yml#L10). The `max_epochs` and `max_steps` params should also be included in `ner_ruler.cfg`

But there is no training with `ner_ruler.cfg` right?

ljvmiranda921:
Yeah, there's no training. But I think the override fails without it. 🤔

kadarakos:
> Yeah, there's no training. But I think the override fails without it. thinking

Sadness okok!

ljvmiranda921:
Ok, I think this is ready for review!

ljvmiranda921:
Scores are much better now. Feel free to add to the rules, @kadarakos 🙇

kadarakos:
> Scores are much better now. Feel free to add to the rules, @kadarakos bow

refactored the opening-hours and added one, hope you can confirm the improvement.

ljvmiranda921:
Ah nice looks good! Can confirm that it works

kadarakos:
> Ah nice looks good! Can confirm that it works

So what is the P, R, F difference now?

ljvmiranda921:
The F-score difference is still low I think, around 0.2pp (76.90 vs. 77.13). I'll come up with additional rules this afternoon and try to push it to at least 1.0pp or 0.80pp (arbitrary) at the least.

ljvmiranda921:
The F-score difference is now around 0.40pp (76.90 vs. 77.38). I'll come up with additional rules  try to push it to at least 0.50pp or 0.80pp (arbitrary) at the least. The README is updated with the current scores.

ljvmiranda921:
The F-score difference is now around 0.40pp (76.90 vs. 77.38). The README is updated with the current scores.

tutorials/spanruler_restaurant_reviews/README.md
```diff
@@ -0,0 +1,113 @@
...
+*Price*, *Dish*, *Amenity*,  and *Cuisine* from restaurant reviews.
+Below are a few examples from the training data:
+
+![](figures/example_00.png)
+![](figures/example_01.png)
+![](figures/example_02.png)
+
+First, we will train an NER-only model and treat it as our baseline. Then, we will
+attach the `SpanRuler` component **before the `ner` component** of the existing
+pipeline. This setup gives us two pipelines we can compare upon.
```
svlandeg, line 23: This is really nice setup and it's one of those very practical considerations that I'm sure will resonate with a lot of people. When we finalize the project, it would be nice to post something about it!

tutorials/spanruler_restaurant_reviews/README.md
```diff
@@ -0,0 +1,113 @@
...
+| Amenity | `master card`, `take credit card`                  | Amenities mention different payment options. |
+| Amenity | `classy`, `clean`                  | Amenities also include adjectives that describe the restaurant. |
+| Location | `in harold square`, `airport`                  | Location also mentions nearby landmarks for a restaurant. |
+
+If we look at the results, we see an increase in performance for the majority
+of entities with rules:
+
+|          | NER only  | With Spanruler  |
+|----------|-----------|-----------------|
+| Price    | **83.72** | 82.90           |
```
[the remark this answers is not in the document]
svlandeg: You could have the one without `Price` be `.v2`, then it's even extremely easy to swap in the config...

tutorials/spanruler_restaurant_reviews/configs/ner_ruler.cfg
```diff
@@ -0,0 +1,47 @@
...
+vectors = "en_core_web_lg"
+init_tok2vec = null
+
+[system]
+gpu_allocator = null
+seed = 0
+
+[nlp]
+lang = "en"
+pipeline = ["tok2vec","ner", "span_ruler"]
```
svlandeg, line 13: Oh wait, is this correct? I thought you said to attach the `SpanRuler` component before the `ner` component, which would also make more sense to me?

tutorials/spanruler_restaurant_reviews/scripts/rules.py
```diff
@@ -0,0 +1,495 @@
+from typing import Dict, List, Any
+from spacy.util import registry
+
+
+Rules = List[Dict[str, Any]]
+
+
+@registry.misc("restaurant_rules.v1")
+def restaurant_span_rules() -> Rules:
```
svlandeg, line 9: The way this is set up, perfectly allows for a `.v2` in which we may keep some of the rules and maybe others not, if they don't improve overal performance.

tutorials/spanruler_restaurant_reviews/test_spanruler_restaurant_reviews.py
```diff
@@ -0,0 +1,9 @@
+from pathlib import Path
+
+from spacy.cli.project.run import project_run
+from spacy.cli.project.assets import project_assets
+
+def test_spanruler_restaurant_reviews():
+    root = Path(__file__).parent
+    project_assets(root)
+    project_run(root, "all", capture=True)
```
svlandeg, line 9: I spy a non-black-formatted file :-)

tutorials/spanruler_restaurant_reviews/patterns.jsonl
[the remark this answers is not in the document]
svlandeg: Yes, I definitely think versioning the rules is the "spaCy v3 way" to go - good catch Ákos!

tutorials/spanruler_restaurant_reviews/configs/ner_ruler.cfg
svlandeg: I've now read some of the other discussions on this thread. Have we experimented with both approaches, compared F-scores between having `span_ruler`  before or after the `ner`?

tutorials/spanruler_restaurant_reviews/README.md
```diff
@@ -0,0 +1,113 @@
...
+| Restaurant_Name| 76.28     | **78.16**       |
+
+Overall, we have better performance for the combined `ner` and `span_ruler`
+pipeline with our set of rules.
+
+|           | NER only | With Spanruler |
+|-----------|----------|----------------|
+| Precision | 77.58    | **78.00**      |
+| Recall    | 76.23    | **76.77**      |
+| F-score   | 76.90    | **77.38**      |
```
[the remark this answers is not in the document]
svlandeg: Just ran this locally and the overal scores I got were similar: 76.27 for the NER-only, 77.04 for NER+spancat.

tutorials/spanruler_restaurant_reviews/README.md
ljvmiranda921: Hi Sofie! Saw the other comment and this one is incorrect. It should be `after`. Thanks for catching this! I think I'll rerun the pipeline again and see which one performs better (adding the SpanRuler before vs. after)? Or would we prefer showing both? 

Although it seems that adding the SpanRuler after resonates with what we want to showcase in this project: SpanRuler as a helper for NER.

tutorials/spanruler_restaurant_reviews/configs/ner_ruler.cfg
ljvmiranda921: I ran a test before and it showed that running the `span_ruler` after `ner` works better (and it seems more consistent with what we want to showcase). However, I did that before adding a few more rules this year (after the holidays). I'll rerun it again and double-check to be sure.

tutorials/spanruler_restaurant_reviews/README.md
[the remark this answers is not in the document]
ljvmiranda921: Ah yeah that definitely makes sense. I can set that one up! Thanks for this advice!

tutorials/spanruler_restaurant_reviews/README.md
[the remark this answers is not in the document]
ljvmiranda921: Sure I can spend some time doing those!

tutorials/spanruler_restaurant_reviews/test_spanruler_restaurant_reviews.py
ljvmiranda921: Ah wow! Thanks for catching!

tutorials/spanruler_restaurant_reviews/configs/ner_ruler.cfg
ljvmiranda921: Reran the pipeline with the `span_ruler` before and it's still not as performant as adding it `after`. Will keep it the same way for now

ljvmiranda921:
Hi @svlandeg ! I updated the rules and I was able to increase the score to 77.62 (vs. 76.90 in the NER-only method). A few changes I did:
- I re-added rules for `Price` that are much more explicit and they managed to increase the score.
- Added a few rules for `Hours`, with positive effect on the score.
- Kept all rules under `v1` instead of creating another ruleset version.

I updated the README and other things! Let me know if the score is passable. If not, I'll spend a bit of time at the end of the week coming up with other rules 🙇

tutorials/spanruler_restaurant_reviews/project.yml
```diff
@@ -0,0 +1,167 @@
...
+      - "python -m spacy convert assets/test.iob corpus/"
+      - "mv corpus/train.spacy corpus/train_dev.spacy"
+      # Split corpus into train, dev, and test
+      - "python -m scripts.split corpus/train_dev.spacy corpus/ --train-size 0.8 --seed ${vars.seed}"
+    deps:
+      - assets/train_raw.iob
+      - assets/test_raw.iob
+    outputs:
+      - corpus/train.spacy
+      - corpus/dev.spacy
```
adrianeboyd, line 112:
```suggestion
      - corpus/dev.spacy
      - corpus/test.spacy
```

tutorials/spanruler_restaurant_reviews/project.yml
```diff
@@ -0,0 +1,167 @@
...
+
+      - >-
+        python -m spacy evaluate
+        models/ner_ruler/
+        corpus/test.spacy
+        --gpu-id ${vars.gpu_id}
+        --output metrics/combined.json
+    deps:
+      - training/ner/model-best/
+      - models/ner_ruler
```
adrianeboyd, line 164:
```suggestion
      - models/ner_ruler
      - corpus/test.spacy
```

adrianeboyd (commented):
Adding `test.spacy` to the deps/output

tutorials/spanruler_restaurant_reviews/.gitignore
```diff
@@ -0,0 +1,168 @@
+assets/
+corpus/
+training/
+metrics/
+models/
+
+
+# Byte-compiled / optimized / DLL files
```
adrianeboyd, line 8: What parts of this `.gitignore` are needed for this particular project subdirectory?

ljvmiranda921: Updated!

tutorials/spanruler_restaurant_reviews/project.yml
```diff
@@ -26,21 +26,21 @@ description: |
...
   |          | NER only  | With Spanruler  |
   |----------|-----------|-----------------|
-  | Price    | 83.72     | **84.88**       |
-  | Rating   | 77.21     | **77.78**       |
-  | Hours    | 64.78     | **65.31**       |
-  | Amenity  | 66.67     | **67.61**       |
-  | Location | 81.17     | **81.99**       |
-  | Restaurant_Name| 76.28     | **78.44**       |
+  | Price    | 81.68     | **83.23**       |
+  | Rating   | 78.42     |   78.06         |
```
kadarakos, line 30: The `Rating` rules are making it worse somehow.

victorialslocum: I'm going to look into this and see if it has to do with annotation inconsistencies!

svlandeg:
@victorialslocum, can you have a look into the failing test?

victorialslocum:
@ljvmiranda921 In order to fix the failing tests, I set up the `ner_ruler.cfg` to point to the original rules instead of the new ones, which made me realize that we're not testing the new workflow in the test file, and that there's really no way to without making separate configs. Is this ok? I've tested the new workflow several times locally, but if we don't end up making another test for it, it might make sense for someone else to test it too. 

Also, should I make a note in the `project.yml` to make sure users change the line in the config before they try the new workflow?

ljvmiranda921:
@victorialslocum Ah I see I get it now why there was a separate config for `restaurant_rules.v2` before. I think it's still nice to have the v2 rules tested in the CI so that everything's covered. 

Perhaps one solution might be:
1. Bring back the config where it uses `restaurant_rules.v2` instead of `v1`
2. Create a new `workflow` that runs all the commands needed to set up the dependencies for v2 to work, so that the new results can be reproduced. You might have already done a  part of this in the `review` workflow. 
3. Then add another test in the test file that runs that new workflow.

While we're at it, it might be nice to rename the `all` workflow into something like `old`, `v1`, `before-review`, etc. and call this new workflow `new`, `v2`, `after-review` (up to you!). 

So for users, they can think of two major workflows: (1) the naive approach of combining rules + models (previously `all`) and the (2) new approach where the annotations were investigated and everything's been rerun. Then there are optional workflows like `prodigy` where they can load in the data and check it themselves (which they don't need to do since we've shipped the post-review JSONL files)

ljvmiranda921:
LGTM!

ljvmiranda921:
LGTM

PR #152, opened 2022-12-14
GoEmotions disclaimer

koaning:
Figured I'd add a disclaimer. It's a great tutorial to show how a CSV works, but it's not a model I'd be comfortable putting into production.

### DIFF
--- a/tutorials/textcat_goemotions/README.md
+++ b/tutorials/textcat_goemotions/README.md
@@ -2,7 +2,10 @@
 
 # 🪐 spaCy Project: Categorization of emotions in Reddit posts (Text Classification)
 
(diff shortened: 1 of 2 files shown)


koaning:
I accidentally used `>` to overwrite the original README. Sorry about that!

adrianeboyd:
I'm not sure all the README generation works as intended in the first place. It seems to clobber when I try it out, so it might not be you.

adrianeboyd:
Ah, fun, `smart-open` broke on Monday. This is fixed in all my various v3.x backports releases, so it will be fixed in spacy very soon.

polm (approved):
Looks good to me, thanks for adding this! 

Additionally, the README generation seems to work fine when I use the `.github/update_docs.py` script.

PR #154, opened 2022-12-19
Span labeling datasets

kadarakos:
Data sets, preprocessing, unseen/seen split and analysis utilities for the technical report.


kadarakos:
The open question that remain:

1. Whether to include the large silver `WikiNeural` dataset. I did not find the results in the original paper encouraging enough for the inclusion.
2. Whether to include `FiNER` since the `spaCy` `ner` pipelines don't really work well on it.
3. Where did the `Toxic Spans` disappear? We do have the `scripts/toxic_spans.py` preprocessing script, but the data set is missing.
4. Should we include the MIT Restaurants since its already used in the `SpanRuler` tutorial? https://github.com/explosion/projects/pull/151

kadarakos:
>     3. Where did the `Toxic Spans` disappear? We do have the `scripts/toxic_spans.py` preprocessing script, but the data set is missing.

Oh, nvm the Toxic Spans is a nice data set, but has a large annotator disagreement, which is interesting, but at this time we are not sure what to do with that.

kadarakos:
> Let's include #4 but leave out the rest?

Thanks for calling it, I think its the right move, I was just sad to have so little amount of data sets in there. I hope we will add a couple more soonish!

ljvmiranda921 (approved):

pmbaumgartner:
On macOS, I get an issue with the `iconv` commands running the `unpack-conll` step. I think this is because `iconv` on macOS doesn't have a `-o` flag. 

In addition, I'm not sure if `iconv` is a dependency Windows users will have. Is there a python-based way to do these encoding conversions?

benchmarks/span-labeling-datasets/project.yml
```diff
@@ -0,0 +1,501 @@
...
+    - "convert-archaeo-spans"
+    - "convert-anem-ents"
+    - "convert-anem-spans"
+    - "preprocess-restaurant"
+    - "convert-restaurant-ents"
+    - "convert-restaurant-spans"
+    - "inspect-restaurant"
+
+assets:
+  # WNUT17 https://aclanthology.org/W17-4418/
```
pmbaumgartner, line 64: I think it would be nice to put these URLs in the `description` field per-asset, that way they get included in the description in the README when you do `spacy project document`, and they aren't hidden in the `project.yml` file.

svlandeg:
> On macOS, I get an issue with the `iconv` commands running the `unpack-conll` step. I think this is because `iconv` on macOS doesn't have a `-o` flag.
> 
> In addition, I'm not sure if `iconv` is a dependency Windows users will have. Is there a python-based way to do these encoding conversions?

Can we set up some tests so this would get caught automatically?

kadarakos:
> On macOS, I get an issue with the `iconv` commands running the `unpack-conll` step. I think this is because `iconv` on macOS doesn't have a `-o` flag.
> 
> In addition, I'm not sure if `iconv` is a dependency Windows users will have. Is there a python-based way to do these encoding conversions?

Replaced the `gzip`, `tar` and `iconv` with `python` functions to be safe. I guess on Windows `mv` doesn't work either?

pmbaumgartner:
> Replaced the `gzip`, `tar` and `iconv` with `python` functions to be safe. I guess on Windows `mv` doesn't work either?

I'm not sure. I usually use a bunch of stuff from [`shutil`](https://docs.python.org/3/library/shutil.html) just to be safe when doing these sort of things.

svlandeg:
Again, we should set up a test that catches this, as the CI runs on both Windows and Linux.

svlandeg:
> Again, we should set up a test that catches this, as the CI runs on both Windows and Linux.

@ljvmiranda921 : is this something you can help look into, with Akos currently on holiday?

ljvmiranda921:
Sure!

ljvmiranda921:
I updated the following:
- Updated the descriptions for each asset, and regenerating the README.md
- Added a test that runs the whole pipeline (it passes now)
- Run formatting tools to the codebase

Let me know if there's anything I should change! 🙇

benchmarks/span-labeling-datasets/scripts/preprocess.py
```diff
@@ -0,0 +1,30 @@
...
+from wasabi import msg
+
+
+def preprocess(
+    input_path: Path, output_path: Path, *, strip_digits: bool = False
+) -> None:
+    """
+    Helper function to canonicalize all datasets into the same ConLL format.
+    """
+    with input_path.open(encoding="utf-8") as f:
```
ljvmiranda921, line 14: Added encoding here so that `UnicodeDecodeError` doesn't occur for Windows. Same goes for file output

benchmarks/span-labeling-datasets/test_span_labeling_datasets.py
```diff
@@ -0,0 +1,10 @@
+from pathlib import Path
+
+from spacy.cli.project.assets import project_assets
+from spacy.cli.project.run import project_run
+
+
+def test_span_labeling_datasets():
```
ljvmiranda921, line 7: The test runs the whole workflow

ljvmiranda921 (commented):
I updated the following:

- Updated the descriptions for each asset, and regenerating the README.md
- Added a test that runs the whole pipeline (it passes now)
- Run formatting tools to the codebase
- Let me know if there's anything I should change! 🙇

PR #157, opened 2022-12-28
Add pull request template

shadeMe:


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,23 @@
...
+<!-- What type of change does your PR cover? Is it a bug fix, an enhancement
+or new feature, or a change to the documentation? -->
+
+## Checklist
+
+<!--- Before you submit the PR, go over this checklist and make sure you can
+tick off all the boxes. [] -> [x] -->
+
+- [ ] I confirm that I have the right to submit this contribution under the project's MIT license.
+- [ ] I ran the tests, and the `update` scripts in the `.github` folder.
```
svlandeg, line 22: I'd have these as two separate boxes, because you typically run the tests before opening the PR, but the "update" script will need to be called right before merging.

Actually, should we have two subsections to make this distinction more clear or is that too much?

shadeMe: Separating them into two checkboxes makes sense, yeah. But what do you mean with subsections?

svlandeg: Never mind that bit 🙈

svlandeg:
Looking good, thanks Madeesh!

PR #158, opened 2022-12-29
Update Python version for CI to 3.9

polm:
## Description

This updates the CI Python version from 3.7 to 3.9 because 3.7 will reach end of life next year. 

### Types of change

chore

## Checklist

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -33,11 +33,19 @@ jobs:
                 versionSpec: '$(python.version)'
                 architecture: 'x64'
 
+
+
           - script: |
                 pip install "spacy>=3.2.0,<3.5.0"
                 pip install pytest
                 pip install wheel
             displayName: 'Install dependencies'
+
+          - if: contains( matrix.imageName, 'macos')
+            name: Install bash 5.0 under macOS for mapfile
+            script: |
+              brew install bash
+              sudo mv /usr/local/bin/bash /bin/bash
 
           - task: Bash@3
             inputs:
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -33,18 +33,16 @@ jobs:
                 versionSpec: '$(python.version)'
                 architecture: 'x64'
 
-
-
           - script: |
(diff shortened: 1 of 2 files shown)


svlandeg (commented):
3.7 still has half a year to go until it reaches EOL, and we still have many users on 3.7, so I wouldn't remove it from the CI entirely. What we can do though, is for instance run 3.7 on Windows and 3.9 on Linux, so we test more broadly.

rmitsch:
Sounds reasonable. Questions:
- Has this been discussed internally yet?
- Why 3.9 and not 3.8?

svlandeg:
We could add MacOS for 3.8 :-)

polm:
I don't think we've discussed this internally. My choice of 3.9 here was basically arbitrary.

I have no objection to having multiple versions of Python in the CI, though before this PR the tests are already just using one version, probably to keep test runtime down. I also do think it's reasonable to expect fewer issues related to Python version here than in the main spaCy repo, for example. 

I'll go ahead and change this to:

- 3.7 Win
- 3.8 OSX
- 3.9 Linux

polm:
Well, this found our first error in the tests: `mapfile` isn't in the default bash on OSX. I'll work on fixing it...

rmitsch:
Which tests did you run to ensure that the behavior of the modified script matches the existing one? Specifically, are tests rerun if changes are detected at the first directory level? Sorry, my bash-fu isn't good enough to see that right away.

polm:
Thanks for mentioning that - I had tested changes in the top-level directory but had forgotten about the README files in the first-level directories, so it would have re-run in that case. I fixed that.

After the fix, the way the bash script works bit by bit is:

```
git diff ... | sed ... # this was already there
cut -f 1-2 -d/ # was being done later anyway, but gives the first two components, and any second slash is removed (there is at most one slash)
grep /. # This finds things with a slash followed by a character (must be a second-level directory)
sort -u # This filters any duplicates, like if `a/b` and `a/b/c` have modifications
```

I tested this by making modifications to files at the top, first, second, and third levels, including duplicates and non-duplicates.

PR #160, opened 2023-01-11
Add automatic doc building

polm:
## Description

This is a script to automatically run update scripts once a week, just like the autoblack job in spaCy.

Currently the config update script (`update_configs.py`) is not included. One reason is that config changes require more review, another is that there may be cases where we don't want configs auto-filled. Another reason is that currently the script just fails with an error on some projects, which needs to be checked separately.

### Types of change

automation

## Checklist

### DIFF
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -5,6 +5,7 @@ on:
   workflow_dispatch:  # allow manual trigger
   schedule:
     - cron: '0 8 * * 5'  # every Friday at 8am UTC
+  push: # TEMP
 
 jobs:
   autodoc:
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -18,6 +18,7 @@ jobs:
       - uses: actions/setup-python@v4
       - name: Build Docs
         run: |
+          pip install spacy
           python .github/update_docs.py
           python .github/update_category_docs.py
           python .github/update_projects_json.py
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -5,7 +5,6 @@ on:
   workflow_dispatch:  # allow manual trigger
   schedule:
     - cron: '0 8 * * 5'  # every Friday at 8am UTC
-  push: # TEMP
 
 jobs:
   autodoc:
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -21,7 +21,8 @@ jobs:
           pip install spacy
           python .github/update_docs.py
           python .github/update_category_docs.py
-          python .github/update_projects_json.py
+          # Temporarily commented out while PR is reviewed
+          # python .github/update_projects_json.py
       - name: Check for modified files
         id: git-check
         run: echo modified=$(if git diff-index --quiet HEAD --; then echo "false"; else echo "true"; fi) >> $GITHUB_OUTPUT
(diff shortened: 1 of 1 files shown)


polm:
Note this runs the json build script in #150, and so shouldn't be merged before that (thus this is a draft). Besides that this PR is complete.

polm:
Checking the config script again, it's not suitable for automation like the other scripts because it can require loading pipelines. This is a problem in some cases such as:

- the pipeline requires a pretrained pipeline like `en_core_web_lg`
- the pipeline requires another build pipeline, like when coref builds on an earlier component

If we can exclude projects somehow then it should be fine though, assuming that automating config changes is desirable.

polm:
#150 is merged, so this is ready.

polm:
Since #150 was reverted, this now depends on #166 for the json script.

rmitsch:
What config script are you referring to - the .jsonl generation script from https://github.com/explosion/projects/pull/166?

polm:
No, the "config script" is `.github/update_configs.py`, which fills configs according to current defaults.

rmitsch:
Ah, I see - sorry for my confusion. Have you triggered this workflow yet?

polm:
No, I haven't triggered this yet. I think I'm supposed to be able to do so from the "Actions" tab on the repo, but because this repo has no existing actions, it shows me a menu of new actions instead of the UI in [this guide](https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow). I think I should be able to do it with the REST API, I can look into that.

rmitsch:
If it's not too much of an effort, testing this with the REST API would be great.

polm:
This was trickier than I expected. It looks like if a workflow has never been run before, it is impossible to see it in the web UI or interact with it through the REST API at all, including triggering runs. The easiest way to trigger it is to add a condition that it runs on every push. If that has been done, rather than interacting with the REST API manually, it's easiest to use the GitHub command line client like this:

```
gh workflow run build-docs.yml --ref chore/auto-update-docs
```

This behaviour is documented at [Stack Overflow](https://stackoverflow.com/questions/63362126/github-actions-how-to-run-a-workflow-created-on-a-non-master-branch-from-the-wo). 

Now the job has been tested ([run log](https://github.com/explosion/projects/actions/runs/3945644240), #169) and seems to be working normally. Since #166 is still in review I have commented that part out for now. If this is reviewed first I could make a separate PR to add in the jsonl generation later.

To avoid excessive autogenerated PRs I will remove the push condition for now.

rmitsch:
Oh, thought it would be more straightforward. Thanks for doing this! Not approving this yet as #166 should be merged first and I don't wanna create confusion, but I will as soon as that one is through.

svlandeg (approved):
Let's get this merged in and see whether it works tomorrow 🤞

polm:
Unfortunately, this failed. 

https://github.com/explosion/projects/actions/runs/4022614279/jobs/6912539633#step:4:107

Copying the error for posterity:

```
/home/runner/work/projects/projects/tutorials/textcat_docs_issues/README.md

✘ The project.yml specifies a spaCy version range (>=3.0.0,<3.5.0)
that's not compatible with the version of spaCy you're running (3.5.0). You can
edit version requirement in the project.yml to load it, but the project may not
run as expected.
```

The issue is that the spaCy version wasn't valid for one of the projects. So things to think about:

- should old version ranges like this be updated, or should we expect them?
- should the action fail in this case or skip the failed project with a note?
- when the action fails, what notifications should be made?

PR #162, opened 2023-01-11
Check all project tests in CI

adrianeboyd:
## Description

I think that #158 only verified that the bash script runs, not that all project tests succeed in the new CI configuration, in particular not for macos.

I also noticed in #161 that the exit code in case of failures doesn't seem to returned as intended even when there are install failures for requirements or pytest failures, so the CI logs need to be inspected manually here.

### Types of change

Tests.

## Checklist

- [ ] I confirm that I have the right to submit this contribution under the project's MIT license.


adrianeboyd:
This is to check all projects with the current setup. It may time out due to the speedster tests (among others), so we may need to increase the timeout in the CI.

PR #163, opened 2023-01-11
Revert #150 to deal with issue in README generation

polm:
This reverts commit bcd8f3bb85ffcb0ba510fda0b4bba02adddc1d8c.

## Description

After merging #150, it was noted that actually having long text in project.yml may be helping to avoid an issue with README generation. This PR reverts #150. 

It may be desirable to revert only the README changes and leave the JSON generation in, since that's technically unrelated.

## Checklist


adrianeboyd:
We double-checked the README generation issue and don't think it's a problem with the current `.github` scripts, but want to split the `README` edits and the `projects.jsonl` changes into two separate PRs.

adrianeboyd:
@polm: Can we revert as a separate PR first?

PR #176, opened 2023-01-27
Remove capture from speedster tests

adrianeboyd:
## Description

Windows tests are failing due `capture=True` and output that's not UTF-8. This is to test whether you can keep `capture=True` if you provide the expected encoding, otherwise it seems like the best option is to use `capture=False`.

### Types of change

Tests.

## Checklist

- [x] I ran the `update` scripts in the `.github` folder, and all the configs and docs are up-to-date.

### DIFF
--- a/experimental/ner_wikiner_speedster/test_project_ner_wikiner_speedster.py
+++ b/experimental/ner_wikiner_speedster/test_project_ner_wikiner_speedster.py
@@ -13,4 +13,4 @@ def test_wikiner_speedster_project():
     }
     root = Path(__file__).parent
     project_assets(root)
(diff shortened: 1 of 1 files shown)


adrianeboyd:
Ah, only an arg to `run_command` and we've hard-coded "utf8" in `run_command`. Nevermind for here, let's try without `capture=True` and we may want to set it to false for windows across the board at this point. (I'm not sure what's going internally in terms of what the instances have as their default encoding, or which libraries/tools have started using different download/progress thingies that lead to this.)

adrianeboyd:
Whatever character it is, it's related to downloading from huggingface and you can see it in the windows logs:

```
Downloading (�)okenizer_config.json:   0%|          | 0.00/29.0 [00:00<?, ?B/s]
```

PR #178, opened 2023-02-01
Open an issue if the auto doc build fails

polm:
## Description

This makes it so that the changes in #160 will open an issue if the build is unable to complete for some reason. So far that has happened due to a conflict in version specifications. While that should also be resolved, one issue is that a failed build may not alert the right people, while an issue is more conspicuous. 

## Checklist


polm:
This approach won't work because this repo doesn't have issues.

PR #197, opened 2023-07-04
Code migration to the newest version of Spacy 3.5

KochevaIvona:
## Types of change
 - Changes in all files where spaCy and depended libraries are used.


svlandeg:
Hi!

It looks like you accidentally committed various unnecessary files, but if I understand it correctly, you're aiming to add the [`nel_wikipedia` project](https://github.com/explosion/projects/tree/master/nel-wikipedia) to the `v3` branch here.

We definitely understand the desire to use the parsing scripts to generate a KB for v3, but we've actually started a more comprehensive project to parse Wikidata/Wikipedia into a `KnowledgeBase`, called [`wikid`](https://github.com/explosion/wikid). This will be easier to maintain and use than the original `nel_wikipedia` project, so we'll favour that going forward. 

So while we appreciate your contribution, we'll go ahead and close this PR. Thanks!

